PR #19, opened 2021-05-21
Updated readme to include a table of badges

abdallahbeshi:


KaliCZ (dismissed):
image
1. Can't we change the text in the badge to just `build & tests`?
2. I would add a first column with the name of the country (so people don't need to read the badge really)
3. Pls add explanation why spain has different version - that it's because of our test certificate
4. Can you make the row headers and title headers bold pls? 😄 

But basically exactly what I wanted.

abdallahbeshi:
> image
> 
> 1. Can't we change the text in the badge to just `build & tests`?
> 2. I would add a first column with the name of the country (so people don't need to read the badge really)
> 3. Pls add explanation why spain has different version - that it's because of our test certificate
> 4. Can you make the row headers and title headers bold pls? 😄
> 
> But basically exactly what I wanted.

1. We can, but then it will be painful to know when action is for which fiscalization when looking at the builds sections in the PR or in github actions.
2. done
3. done
4. done

KaliCZ:
If changing the text means changing the name of the build, then the answer to `can't we just change the text in the badge?` is no 😄

KaliCZ:
I made the project name into a first column and also shortened the badge definition by a lot as per - https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge

Dunno why you had the link there twice eveytime 🤷

KaliCZ (approved):

PR #23, opened 2021-05-24
Introduced publish for all fiscalizations.

abdallahbeshi:


KaliCZ (approved):
Let's make the publish only run if the tests pass.

PR #32, opened 2021-05-26
Updated Hungarian fiscalization to V3.0 part 1

abdallahbeshi:
Tests are failing because our test account is misconfigured and in order to fix that we need access to the test environment UI which is down at the moment.

There will be another PR that will contain more test cases.


KaliCZ (commented):
And then all of those tests need to be rewritten when we make the classes strongly-typed, because you can no longer use TestCase with enum values.

src/Hungary/Mews.Fiscalizations.Hungary/Models/TaxPayer/TaxPayerData.cs
```diff
@@ -4,12 +4,13 @@ namespace Mews.Fiscalizations.Hungary.Models
 {
     public sealed class TaxPayerData
```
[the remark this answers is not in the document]
abdallahbeshi: Nope, the endpoint we use for this is to query a tax id and the government would return us some information about that tax id.

so it could be anyone's tax id.

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
```diff
@@ -38,14 +44,33 @@ internal static Dto.ManageInvoiceRequest CreateManageInvoicesRequest(TechnicalUs
...
+                    invoiceData = invoiceDataBytes,
+                    invoiceOperation = operation,
+                    electronicInvoiceHash = invoiceData.completenessIndicator.Match(
+                        t => new Dto.CryptoType
+                        {
+                            cryptoType = RequestEncryptionAlgorithm,
+                            Value = Sha512.GetSha3Hash(Convert.ToBase64String(invoiceDataBytes))
+                        },
+                        f => null
+                    )
```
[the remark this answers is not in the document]
abdallahbeshi: image

This has nothing to do with if the receiver is a company or not.
completenessIndicator indicates whether the data report is the invoice itself (no more data on the invoice), so i think this should be true for everything except for private persons as they don't provide any info like (tax id, address or name).

src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
```diff
@@ -164,6 +200,15 @@ private static Dto.LineAmountsNormalType MapLineAmounts(InvoiceItem item)
             };
         }
 
+        private static Dto.CustomerVatStatusType MapCustomerVatStatusType(CustomerVatStatusType type)
+        {
+            return type.Match(
+                CustomerVatStatusType.Domestic, _ => Dto.CustomerVatStatusType.DOMESTIC,
```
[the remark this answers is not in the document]
abdallahbeshi: Based on my understanding, this is not really used to identify if the receiver is a customer or a company
image

we can take a look at the documentation together and confirm

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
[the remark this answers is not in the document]
KaliCZ: Aha :D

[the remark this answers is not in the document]
KaliCZ: Well that assumption `so i think this should be true for everything except for private persons` should be verified by Martin.

src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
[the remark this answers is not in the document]
KaliCZ: Private person = natural person = a human being. The others are considered companies.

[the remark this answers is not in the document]
abdallahbeshi: I deleted the ENUM as its no longer needed (we can use the coproduct to specify which one to send).

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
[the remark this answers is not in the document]
abdallahbeshi: image
So, it should be always true, as we are reporting ELECTRONIC invoices.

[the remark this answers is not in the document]
abdallahbeshi: image
So for private_persons it should be false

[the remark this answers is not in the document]
KaliCZ: What about the third one. Are we issuing electronic invoices?

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/Company.cs
```diff
@@ -0,0 +1,17 @@
+﻿using FuncSharp;
+
+namespace Mews.Fiscalizations.Hungary.Models
+{
+    public sealed class Company : Coproduct2<LocalCompany, ForeignCompany>
```
KaliCZ, line 5: Do we actually need the type for this? I wanted to just check the country in the serialization. We'd need to decide if address or the identifier country is the deciding factor. But technically we're deciding that here in the static create anyway - you can't create a LocalCompany if the taxId isn't hungarian, not the address.

KaliCZ (commented):
We'll review the tests after the test environment comes back up :)

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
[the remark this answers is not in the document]
abdallahbeshi: Yes, we only issue electronic invoices.

src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
```diff
@@ -100,6 +125,43 @@ private static Dto.InvoiceType GetCommonInvoice(Invoice invoice, IEnumerable<Dto
 
         }
 
+        private static IOption<Dto.CustomerVatDataType> GetCustomerVatDataType(Receiver receiver)
+        {
+            return receiver.Match(
+                customer => Option.Empty<Dto.CustomerVatDataType>(),
+                company => company.Match(
+                    local => GetCustomerVatDataType(local.TaxpayerId).ToOption(),
+                    foreign => foreign.TaxpayerId.Map(i => GetCustomerVatDataType(i))
```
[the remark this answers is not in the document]
abdallahbeshi: If there is no taxId provided, then we shouldn't fill the `customerVatData` because it all depends on the tax id.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ModificationInvoice.cs
```diff
@@ -30,13 +30,13 @@ public sealed class ModificationInvoice : Invoice
             DateTime issueDate,
             DateTime paymentDate,
             SupplierInfo supplierInfo,
```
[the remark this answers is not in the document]
abdallahbeshi: Yea i noticed them but i didn't want to make this PR huge, this can wait, so i will revisit it when we're done with the upgrade to V3

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/Company.cs
abdallahbeshi: I would keep it just because it makes the code more readable, setting values based on the data provided would seem like some magic.

src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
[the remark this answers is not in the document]
KaliCZ: I'm just poking into the possibility that the server will decline the invoice because we didn't put any info about the receiver tax info

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
```diff
@@ -21,11 +21,24 @@ private ForeignCompany(Name name, SimpleAddress address, TaxpayerIdentificationN
...
+                    _ => Try.Success<ForeignCompany, IEnumerable<Error>>(new ForeignCompany(n, a))
+                )
             );
+
+            return result.FlatMap(r => r);
+        }
+
+        private static bool IsForeignTaxpayerNumber(TaxpayerIdentificationNumber number)
+        {
+            return !number.Country.Alpha2Code.Equals(Countries.Hungary.Alpha2Code);
```
KaliCZ, line 41: We have equality on countries, right? You don't need to compare on AlphaCode, you can compare countries directly.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
```diff
@@ -21,11 +21,24 @@ private ForeignCompany(Name name, SimpleAddress address, TaxpayerIdentificationN
...
                 ObjectValidations.NotNull(name),
                 ObjectValidations.NotNull(address),
-                (n, a) => new ForeignCompany(n, a, taxpayerId)
+                (n, a) => taxpayerId.ToOption().Match(
+                    i => IsForeignTaxpayerNumber(i).ToTry(
+                        t => new ForeignCompany(n, a, i),
+                        f => new Error($"{nameof(TaxpayerIdentificationNumber)} must be a foreign (non-Hungarian) taxpayer number.").ToEnumerable()
+                    ),
+                    _ => Try.Success<ForeignCompany, IEnumerable<Error>>(new ForeignCompany(n, a))
+                )
```
KaliCZ, line 33: Just do 
var id = taxpayerId.ToOption().Where(i => IsForeignTaxpayerNumber(i)).ToTry(....)

And then put it into the try aggregate as a third Try, so you don't need to flatmap.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
```diff
@@ -21,11 +21,24 @@ private ForeignCompany(Name name, SimpleAddress address, TaxpayerIdentificationN
...
             );
+
+            return result.FlatMap(r => r);
+        }
+
+        private static bool IsForeignTaxpayerNumber(TaxpayerIdentificationNumber number)
```
KaliCZ, line 39: I think that after you do the 2 comment ideas, this method will not be needed as it will be a simple inlined variable.

It's easier to name a variable rather than create a function just so that the condition has a name.

KaliCZ (approved):
Yeah, basically gut now.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
abdallahbeshi: Nope we don't, i have a reminder to change it in a separate PR and fix the usages.

KaliCZ: Well we have a list of countries, so you can't actually create your own hungary. You can do GetByCode. So you're getting the same hungary everytime. So Equality through reference equality should still work.

KaliCZ (dismissed):

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
```diff
@@ -21,12 +21,12 @@ private ForeignCompany(Name name, SimpleAddress address, TaxpayerIdentificationN
 
         public static ITry<ForeignCompany, IEnumerable<Error>> Create(Name name, SimpleAddress address, TaxpayerIdentificationNumber taxpayerId = null)
         {
-            var foreignTaxPayerId = taxpayerId.ToOption().Where(i => !i.Country.Equals(Countries.Hungary));
+            var foreignTaxPayerId = taxpayerId.ToOption().ToOption().Where(i => i.Map(n => !n.Country.Equals(Countries.Hungary)).GetOrElse(true));
```
KaliCZ, line 24: Can we maybe make this multiline to show what it does in a more explicit way?
```
var optionalForeignTaxPayerId = taxpayerId.ToOption().ToOption().Where(i => i.Match(
    identifier => identifier.Country.Equals(Countries.Hungary)
    _ => true
);
```

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/Company.cs
marektresnak: I'd keep the two classes. I like that you can't create LocalCompany without TaxPayerId.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
```diff
@@ -0,0 +1,36 @@
...
+        public static ITry<ForeignCompany, IEnumerable<Error>> Create(Name name, SimpleAddress address, TaxpayerIdentificationNumber taxpayerId = null)
+        {
+            var optionalForeignTaxPayerId = taxpayerId.ToOption().ToOption().Where(i => i.Match(
+                identifier => identifier.Country.Equals(Countries.Hungary),
+                _ => true
+            ));
+            return Try.Aggregate(
+                optionalForeignTaxPayerId.ToTry(_ => Error.Create($"{nameof(TaxpayerIdentificationNumber)} must be a foreign (non-Hungarian) taxpayer number.")),
+                ObjectValidations.NotNull(name),
+                ObjectValidations.NotNull(address),
```
marektresnak, line 31: Didn't we say we will avoid these `NotNull` validations because they are not practical? I guess it doesn't really matter if there are other validations and the return type is already an ITry, but it can simplify methods that would otherwise have a simple return type. But overall, I think it should be consistent everywhere.

cc @KaliCZ

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
```diff
@@ -61,25 +86,34 @@ private static Dto.ManageInvoiceRequest CreateManageInvoicesRequest<T>(Technical
         }
 
         private static T CreateRequest<T>(TechnicalUser user, SoftwareIdentification software, string additionalSignatureData = null)
-            where T : Dto.BasicRequestType, new()
+            where T : Dto.BasicOnlineInvoiceRequestType, new()
         {
             var requestId = RequestId.CreateRandom();
-            var timestamp = DateTime.UtcNow;
+            var nowUtc = DateTime.UtcNow;
+            var timestamp = new DateTime(nowUtc.Year, nowUtc.Month, nowUtc.Day, nowUtc.Hour, nowUtc.Minute, nowUtc.Second, DateTimeKind.Utc);
```
marektresnak, line 93: I suppose this is about dropping the information about milliseconds?

src/Hungary/Mews.Fiscalizations.Hungary.Tests/StatusTests.cs
```diff
@@ -7,15 +9,37 @@ namespace Mews.Fiscalizations.Hungary.Tests
...
-            Assert.IsNull(status.GeneralErrorResult);
+        [Test]
+        public async Task GetExchangeTokenSucceeds()
+        {
+            var navClient = TestFixture.GetNavClient();
+            var exchangeToken = await navClient.GetExchangeTokenAsync();
+            AssertResponse(exchangeToken);
+        }
+
+        private void AssertResponse<TResult, TCode>(ResponseResult<TResult, TCode> responseResult)
```
marektresnak, line 36: Use `TestFixture.AssertResponse` instead of this?

src/Hungary/Mews.Fiscalizations.Hungary.Tests/InvoiceTests.cs
```diff
@@ -1,184 +1,336 @@
...
+                token: exchangeToken.SuccessResult,
+                invoices: Sequence.FromPreordered(new[] { invoice }, startIndex: 1).Get()
+            );
+
+            Thread.Sleep(2000);
+
+            var sendInvoiceTransactionStatus = await navClient.GetTransactionStatusAsync(sendInvoiceResponse.SuccessResult);
+            TestFixture.AssertResponse(sendInvoiceTransactionStatus);
+            AssertInvoiceStatuses(sendInvoiceTransactionStatus.SuccessResult.InvoiceStatuses);
+
```
marektresnak, line 178: There's quite a lot of code duplication. Could we just have a parametrized helper for this?

src/Hungary/Mews.Fiscalizations.Hungary/RequestCreator.cs
abdallahbeshi: Yup, otherwise it would fail :D they have this validation on their end.

src/Hungary/Mews.Fiscalizations.Hungary.Tests/InvoiceTests.cs
abdallahbeshi: As discussed, I am planning to refactor the tests later when the NAV test UI is back online.

marektresnak (approved):

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/ForeignCompany.cs
KaliCZ: We said that we will not write the Create methods returning an ITry if all it is checking are nulls. But if we already have a create method, we will check the nulls, because if we have a method returning an ITry, it should be complete.

PR #36, opened 2021-05-31
Fixed project dependencies

abdallahbeshi:


KaliCZ (commented):
I'm thinking about it. This will make the users install BouncyCastle without knowing about it. That's why Mews doesn't build now after getting the All package.

The package should rather reference BouncyCastle to make it known to the users that it's required. But do we want to link all the required dependencies like that?

PR #48, opened 2021-06-30
Migrated Czech fiscalization tests to NUnit and fixed an issue with creating a the signedXML

abdallahbeshi:
CZ fiscalization tests never actually got executed/triggered, that's why we never knew about the issue, so i migrated the tests to use NUnit and fixed the issue with signing the xml.

### DIFF
--- a/src/Czechia/Mews.Fiscalizations.Czechia/Communication/SoapSigner.cs
+++ b/src/Czechia/Mews.Fiscalizations.Czechia/Communication/SoapSigner.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
+using System.Deployment.Internal.CodeSigning;
+using System.Security.Cryptography;
 using System.Security.Cryptography.Xml;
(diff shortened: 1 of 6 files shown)


src/Czechia/Mews.Fiscalizations.Czechia/Communication/SoapSigner.cs
```diff
@@ -32,6 +34,7 @@ public SoapSigner(Certificate certificate, SignAlgorithm signAlgorithm)
 
         public XmlDocument SignMessage(XmlDocument xmlDoc)
         {
+            CryptoConfig.AddAlgorithm(typeof(RSAPKCS1SHA256SignatureDescription), "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256");
```
marektresnak, line 37: It would be nicer if we could manage it without this global setting. But after some quick searching, I also didn't find a solution.

marektresnak (approved):

PR #51, opened 2021-07-08
Introduced GetTransaction

abdallahbeshi:


src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -84,6 +84,18 @@ public async Task<ResponseResult<Tss>> UpdateTssAsync(AccessToken token, Guid ts
...
+        {
+            var request = RequestCreator.GetTransaction(tssId, transactionId);
+            return await Client.ProcessRequestAsync<Dto.GetTransactionRequest, Dto.TransactionResponse, Transaction>(
+                method: HttpMethod.Get,
+                endpoint: $"tss/{tssId}/tx/{transactionId}",
+                request: request,
+                successFunc: response => ModelMapper.MapTransaction(response),
+                token: token
+            ).ConfigureAwait(continueOnCapturedContext: false);
+        }
```
KaliCZ, line 97: Or we could just be returning the task directly. It does the same and saves some code. But I guess I would wait with that till after @JakubLinhart and @starychfojtu come up with some strategy on how to write async code consistently.

```suggestion
        public Task<ResponseResult<Transaction>> GetTransactionAsync(AccessToken token, Guid tssId, Guid transactionId)
        {
            var request = RequestCreator.GetTransaction(tssId, transactionId);
            return Client.ProcessRequestAsync<Dto.GetTransactionRequest, Dto.TransactionResponse, Transaction>(
                method: HttpMethod.Get,
                endpoint: $"tss/{tssId}/tx/{transactionId}",
                request: request,
                successFunc: response => ModelMapper.MapTransaction(response),
                token: token
            );
        }
```

KaliCZ (dismissed):

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
JakubLinhart: Yes, there is no need for async/await and you can return task directly. Unfortunately async/await strategy is not in scope of this tertile. Hopefully we will continue discussing it soon.

abdallahbeshi: Yea, we actually should go through the fiscalization libraries and improve it.

I wouldn't block this PR for this, but we will definitely pay it a visit at some point and fix all of them.

KaliCZ (approved):

PR #53, opened 2021-07-12
Added paymentMethod to HU fiscalization.

abdallahbeshi:


src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
```diff
@@ -69,8 +69,11 @@ private static Dto.InvoiceType GetCommonInvoice(Invoice invoice, IEnumerable<Dto
                         invoiceCategory = Dto.InvoiceCategoryType.AGGREGATE,
                         invoiceDeliveryDate = invoice.DeliveryDate,
                         paymentDate = invoice.PaymentDate,
+                        paymentDateSpecified = true,
                         selfBillingIndicator = invoice.IsSelfBilling,
-                        cashAccountingIndicator = invoice.IsCashAccounting
+                        cashAccountingIndicator = invoice.IsCashAccounting,
+                        paymentMethod = invoice.PaymentMethod.Map(m => MapPaymentMethod(m)).GetOrElse(Dto.PaymentMethodType.OTHER),
```
KaliCZ, line 75: I'd rather send null if we don't have it specified. Sending Other is a business decision that changes our logic. Sending null is the same as we've had until now.

KaliCZ: Alternatively we need to wait for Josef to confirm.

KaliCZ: You're also setting `paymentMethodSpecified` to false in case the option is empty so I would expect it to not read the value anyway :) That's the default behavior of XML serialization. The boolean tells you whether to check the value.

KaliCZ (approved):

src/Hungary/Mews.Fiscalizations.Hungary/RequestMapper.cs
abdallahbeshi: As discussed, the `paymentMethod `is not nullable, and in case the value wasn't provided, we will set `paymentMethodSpecified `to false which will ignore the value of `paymentMethod`

PR #56, opened 2021-07-16
Refactored the German fiscalization async code

abdallahbeshi:


src/Germany/Mews.Fiscalizations.Germany/Client.cs
```diff
@@ -26,7 +26,7 @@ internal Client()
             where TResult : class
         {
             var httpResponse = await SendRequestAsync(method, endpoint, request, token).ConfigureAwait(continueOnCapturedContext: false);
-            return await DeserializeAsync(httpResponse, successFunc);
+            return await DeserializeAsync(httpResponse, successFunc).ConfigureAwait(continueOnCapturedContext: false);
```
KaliCZ, line 29: Or we could just do this (everywhere). I know we didn't yet tackle this in a systematic way, but can we at least agree on not awaiting something for no reason at all? @starychfojtu @JakubLinhart 
```suggestion
            return DeserializeAsync(httpResponse, successFunc);
```

src/Germany/Mews.Fiscalizations.Germany/Client.cs
```diff
@@ -36,28 +36,30 @@ internal Client()
             HttpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token.Value);
             var uri = new Uri(BaseUri, $"{RelativeApiUrl}{endpoint}");
             var httpResponse = await HttpClient.GetAsync(uri).ConfigureAwait(continueOnCapturedContext: false);
-            return await DeserializeAsync(httpResponse, successFunc);
+            return await DeserializeAsync(httpResponse, successFunc).ConfigureAwait(continueOnCapturedContext: false);
```
KaliCZ, line 39: same here

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -47,81 +47,76 @@ public async Task<ResponseResult<Model.Client>> CreateClientAsync(AccessToken to
             ).ConfigureAwait(continueOnCapturedContext: false);
```
KaliCZ, line 47: same here. ANd everywhere, I won't go method by method , commenting the same. 😅
But let's wait for the answer I guess.

src/Germany/Mews.Fiscalizations.Germany/Client.cs
abdallahbeshi: The method is declared async, so I believe it's not a good practice to not use await in some places (for example in this case to deserialize the response).
and this method is declared async because the code is awaiting for a response a few lines above. (SendRequestAsync)

KaliCZ: My bad. In this case the compiler doesn't let you not await the result as we already await above. The map would be a good option if available to us as we wouldn't need to await.

KaliCZ (approved):

marektresnak (approved):

PR #59, opened 2021-07-20
Refactored async code

abdallahbeshi:
.NET Core does not have a `SynchronizationContext`, so it doesn't matter if we call ConfigureAwait, so we're removing it from the codebase, please be aware that this can cause deadlocks if you're still using .net framework.


oleg-sokolansky (approved):

src/Czechia/Mews.Fiscalizations.Czechia.Tests/IntegrationTests/Basics.cs
```diff
@@ -30,7 +30,7 @@ public void TimeoutWorks()
             var certificate = CreateCertificate(Fixtures.Second);
             var record = CreateSimpleRecord(certificate, Fixtures.Second);
             var client = new EetClient(certificate, EetEnvironment.Playground, TimeSpan.FromMilliseconds(1));
-            Assert.ThrowsAsync<TaskCanceledException>(async () => await client.SendRevenueAsync(record));
+            Assert.ThrowsAsync<TaskCanceledException>(async () => await client.SendRevenueAsync(record).ConfigureAwait(continueOnCapturedContext: false));
```
KaliCZ, line 33: But here this should work. 
```suggestion
            Assert.ThrowsAsync<TaskCanceledException>(() => client.SendRevenueAsync(record));
```

src/Czechia/Mews.Fiscalizations.Czechia.Tests/IntegrationTests/Basics.cs
```diff
@@ -105,7 +105,7 @@ public void LoggingIsSerializable()
                     StringAssert.StartsWith("{", jsonString);
                 })
             );
-            Assert.DoesNotThrowAsync(async () => await client.SendRevenueAsync(record));
+            Assert.DoesNotThrowAsync(async () => await client.SendRevenueAsync(record).ConfigureAwait(continueOnCapturedContext: false));
```
KaliCZ, line 108: And here.

src/Hungary/Mews.Fiscalizations.Hungary.Tests/InvoiceTests.cs
```diff
@@ -95,29 +95,29 @@ public async Task SendCorrectionForeignCompanyInvoiceSucceeds(string countryCode
...
 
         private async Task<ResponseResult<string, ResultErrorCode>> SendModificationInvoices(Receiver receiver, InvoiceNumber originalInvoiceNumber)
         {
-            var exchangeToken = await NavClient.GetExchangeTokenAsync();
+            var exchangeToken = await NavClient.GetExchangeTokenAsync().ConfigureAwait(continueOnCapturedContext: false);
             return await NavClient.SendModificationDocumentsAsync(
                 token: exchangeToken.SuccessResult,
                 invoices: Sequence.FromPreordered(new[] { CreateModificationInvoice(originalInvoiceNumber, receiver) }, startIndex: 1).Get()
-            );
+            ).ConfigureAwait(continueOnCapturedContext: false);
```
KaliCZ, line 120: Meh, this is starting to be a lot of code. Maybe let's rethink it and talk about it tomorrow if we actually want to keep targetting .net framework 😄 
cc @marektresnak

marektresnak: Maybe we actually could. I don't think we have many users that are writing blocking code on the old platform (if any). So if we just release it as a new major version to make it clear there might be compatibility issues, it shouldn't really affect anyone.

My guess is that only the Czech library is actually being used by other users, and as far as I remember, we haven't done any significant changes in a long time (just some with the migration to .NET core). So not upgrading to the new version is still an option for anyone, who requires `ConfigureAwait`.

KaliCZ (dismissed):

KaliCZ (commented):
What about the `All`.csproj ?

KaliCZ (approved):

PR #62, opened 2021-07-21
Improved the German fiscalization readme

abdallahbeshi:
image

### DIFF
--- a/src/Germany/README.md
+++ b/src/Germany/README.md
@@ -4,7 +4,7 @@
     </a>
     <br><br>
     <b>Mews.Fiscalizations.Germany</b> is a .NET library that was built to help reporting of e-invoices to the German authorities (BSI - Bundesamt für Sicherheit in der Informationstechnik) using [Fiskaly KassenSichV API V1](https://developer.fiskaly.com/api/kassensichv/v1/).
-    <br>
+    <br><br>
     <a href="https://www.nuget.org/packages/Mews.Fiscalizations.Germany/">
         <img src="https://img.shields.io/nuget/v/Mews.Fiscalizations.Germany">
     </a>
--- a/src/Germany/README.md
+++ b/src/Germany/README.md
@@ -1,18 +1,49 @@
-# Mews.Fiscalizations.Germany
+<p align="center">
+    <a href="https://gofiber.io">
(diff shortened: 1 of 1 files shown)


src/Germany/README.md
```diff
@@ -1,18 +1,50 @@
...
+-   All endpoints are covered with tests.
+-   Intuitive immutable DTOs.
+-   Pipelines that run on both Windows and Linux operating systems.
+-   Cross platform (uses .NET Standard).
 
-## Usage
-We tend to use immutable DTOs wherever possible, especially to ensure data validity.
-We want the library to throw an error as soon as possible, i.e. when constructing corresponding data structures.
-That is why we even introduce wrappers for simple datatypes.
+## 👀 Code Examples
```
KaliCZ, line 45: If we keep this, then let's make it the last section (so After the Nuget section)

But actually, I'm thinking about creating a private class Code Examples. With methods named as their use-cases.

So there would be a method `Creating client ` with an XML comment saying what you wrote here. `Fiskaly Client can be created using the **ApiKey** and **ApiSecret** which can be created through Fiskaly dashboard.`

And here we would only add link to the file with the examples. This makes sure that the code example is always up-to-date as it needs to compile. What do we think? cc @marektresnak

KaliCZ (approved):

src/Germany/README.md
abdallahbeshi: I think the tests should be enough, and we mentioned that for more examples, check the tests.

Having code examples in the codebase will require us to modify the tests + the examples if we change the code.

so at the end, we will always have to make 2 changes when modifying the code
either the tests + the code examples in the documentation
or
the tests + the code examples in the codebase.

Most of the github repos i visited, like 90% have code examples in the readme.


Also, thinking about this, if one day i was looking for a similar library, I will be interested in 4 things
1. How to create the client (credentials, certificate...etc)
2. How the request looks like
3. How the response looks like
4. The name of the endpoint and what it does.

KaliCZ: Yeah, but when we actually modify it, we will not remember to update the code example. That's why I would prefer it inside a class. 

But as for now, it's a good improvement, so I'm merging. I'm just saying IMO having a class for that would be helpful.

KaliCZ (approved):

PR #69, opened 2021-07-26
Introduced CODE_OF_CONDUCT.md

abdallahbeshi:


KaliCZ (approved):
I mean this has basically no value to us, but we indeed don't discriminate, so why the hell not include it 🤷 

Also, I know it's a copy paste from somewhere, but can we actually ban somebody? That whole section seems like we can't apply it.

marektresnak (approved):
Hopefully something that we'll never need 😄

PR #76, opened 2021-09-22
Upgraded the library to use the new Fiskaly certified API v2.

abdallahbeshi:
## Description

Upgraded the library to use the new Fiskaly certified API v2.

Fixes #RND-77090

## Type of change
- [ ] Bug fix.
- [x] Feature.
- [ ] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.


src/Germany/Mews.Fiscalizations.Germany.Tests/TransactionTests.cs
```diff
@@ -29,7 +29,7 @@ public async Task GetTransactionSucceeds()
             var retrievedStartedTransaction = await client.GetTransactionAsync(accessToken.SuccessResult, TestFixture.TssId, startedTransaction.SuccessResult.Id);
             Assert.AreEqual(retrievedStartedTransaction.SuccessResult.State, TransactionState.Active);
 
-            var finishedTransaction = await client.FinishTransactionAsync(accessToken.SuccessResult, TestFixture.ClientId, TestFixture.TssId, GetBill(), startedTransaction.SuccessResult.Id, lastRevision: "1");
+            var finishedTransaction = await client.FinishTransactionAsync(accessToken.SuccessResult, TestFixture.ClientId, TestFixture.TssId, GetBill(), startedTransaction.SuccessResult.Id, lastRevision: "2");
```
KaliCZ, line 32: What does this even mean? 😄

src/Germany/Mews.Fiscalizations.Germany/RequestCreator.cs
```diff
@@ -74,6 +73,23 @@ internal static Dto.AuthorizationTokenRequest CreateAuthorizationToken(ApiKey ap
             };
         }
 
+        internal static Dto.AdminLoginRequest CreateAdminLoginRequest(string adminPin)
```
KaliCZ, line 76: I wouldn't prefix this with Create. Just `AdminLoginRequest`. All the other `Create` words here mean that we are actually creating something. But you're not creating a login. A login isn't an entity.

src/Germany/Mews.Fiscalizations.Germany/Model/Result/ResponseResult.cs
```diff
@@ -15,10 +17,7 @@ internal ResponseResult(TResult successResult = null, ErrorResult errorResult =
 
         public bool IsSuccess
         {
-            get
-            {
-                return SuccessResult != null;
-            }
+            get { return ErrorResult.IsNull(); }
```
KaliCZ, line 20: There's a story behind this change. Where it can happen that both results are null.

src/Germany/Mews.Fiscalizations.Germany/Dto/AdminPinRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Newtonsoft.Json;
+
+namespace Mews.Fiscalizations.Germany.Dto
+{
+    internal sealed class AdminPinRequest
```
KaliCZ, line 5: Like this?
```suggestion
    internal sealed class AdminSetPinRequest
```

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -32,16 +31,22 @@ public Task<ResponseResult<Model.Client>> GetClientAsync(AccessToken token, Guid
 
         public async Task<ResponseResult<Model.Client>> CreateClientAsync(AccessToken token, Guid tssId)
         {
-            var tss = await GetTssAsync(token, tssId);
-            var tssCertificate = tss.SuccessResult.Certificate;
-            var certificate = new X509Certificate2(Encoding.UTF8.GetBytes(tssCertificate));
-            var serialNumber = certificate.GetCertHashString();
-
-            var request = RequestCreator.CreateClient(serialNumber);
+            var id = Guid.NewGuid();
```
KaliCZ, line 34: So looks like this should be an optional parameter?

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -56,21 +61,21 @@ public Task<ResponseResult<Tss>> GetTssAsync(AccessToken token, Guid tssId)
...
             );
         }
 
-        public Task<ResponseResult<Tss>> UpdateTssAsync(AccessToken token, Guid tssId, TssState state)
+        public async Task<ResponseResult<Tss>> UpdateTssAsync(AccessToken token, Guid tssId, TssState state)
         {
-            return Client.ProcessRequestAsync<Dto.UpdateTssRequest, Dto.TssResponse, Tss>(
-                method: HttpMethod.Put,
+            return await Client.ProcessRequestAsync<Dto.UpdateTssRequest, Dto.TssResponse, Tss>(
+                method: new HttpMethod("PATCH"),
```
KaliCZ, line 78: `HttpMethod.Patch` doesn't exist?

abdallahbeshi: Not in .Net Standard 2.0

if we update to .NET Standard 2.1, then we can do HttpMethod.Patch

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
abdallahbeshi: image

KaliCZ: Not really sure what you're trying to say? 
I'm saying that the code that wants to create a client should be capable of defining the ID of the client. Instead of you generating a random one.

src/Germany/Mews.Fiscalizations.Germany.Tests/TransactionTests.cs
abdallahbeshi: The initial transaction must start with revision 1, and each update on the transaction must increment it by 1.

src/Germany/Mews.Fiscalizations.Germany/Model/Result/ResponseResult.cs
abdallahbeshi: When we get nothing in the response (example: Login/Logout)

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
abdallahbeshi: I was trying to answer your question but seems like i misunderstood it.

Yes, this can be optional (anyway we return the ID in the response).

abdallahbeshi: it just looks weird that u create something and u have to provide the ID for it 😄 
and since we return the ID in the response, then there is no benefit from adding the ID in the param as nullable.

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -102,7 +107,7 @@ public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken toke
         {
             return Client.ProcessRequestAsync<Dto.FinishTransactionRequest, Dto.TransactionResponse, Transaction>(
                 method: HttpMethod.Put,
-                endpoint: $"tss/{tssId}/tx/{transactionId}?last_revision={lastRevision}",
+                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision={lastRevision}",
```
marektresnak, line 110: So now we're passing new revision instead of previous revision? If so, we should also rename `lastRevision` parameter.

src/Germany/Mews.Fiscalizations.Germany/Model/CreateTss.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Mews.Fiscalizations.Germany.Model
+{
+    public sealed class CreateTss : Tss
```
marektresnak, line 3: Nit: CreatedTss or CreateTssResult?

src/Germany/Mews.Fiscalizations.Germany/Model/Transaction.cs
```diff
@@ -26,6 +34,14 @@ public sealed class Transaction
 
         public Guid Id { get; }
 
+        public Guid ClientId { get; }
+
+        public Guid TssId { get; }
+
+        public string ClientSerialNumber { get; }
+
+        public string TssSerialNumber { get; }
```
marektresnak, line 43: Is CertificateSerial useful after adding these?

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -30,18 +29,24 @@ public Task<ResponseResult<Model.Client>> GetClientAsync(AccessToken token, Guid
...
-            var serialNumber = certificate.GetCertHashString();
-
-            var request = RequestCreator.CreateClient(serialNumber);
+            var clientId = id ?? Guid.NewGuid();
             return await Client.ProcessRequestAsync<Dto.CreateClientRequest, Dto.ClientResponse, Model.Client>(
                 method: HttpMethod.Put,
-                endpoint: $"tss/{tssId}/client/{Guid.NewGuid()}",
-                request: request,
+                endpoint: $"tss/{tssId}/client/{clientId}",
+                request: RequestCreator.CreateClient($"ERS {clientId}"),
```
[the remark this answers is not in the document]
abdallahbeshi: image
This is a postman configuration created by fiskaly and it has examples of each endpoint.

also, this exists in their documentation, but not sure if its mandatory.

src/Germany/Mews.Fiscalizations.Germany/Model/Transaction.cs
abdallahbeshi: I would keep it till we confirm what actually should be displayed on the bill.

Then i can revisit this and delete anything we're not showing on the invoice.

marektresnak: If the API is still providing this value, I'd definitely keep it. The lib should be usable for wider audience, so there's no reason to delete the properties just because we are not using them.

My main point was that maybe this was replaced by the new serial numbers. But if that's not the case, let's keep it the way it is.

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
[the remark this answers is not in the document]
marektresnak: I would say it's up to us to actually choose this. Therefore, it should be up to the user of the library to specify the serial number.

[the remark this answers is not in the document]
abdallahbeshi: Yea, i checked again and you're right.

src/Germany/Mews.Fiscalizations.Germany/Model/Transaction.cs
abdallahbeshi: Nope, i don't think anything replaces anything.

src/Germany/Mews.Fiscalizations.Germany/ModelMapper.cs
```diff
@@ -64,22 +68,52 @@ internal static ResponseResult<Tss> MapTss(Dto.TssResponse tss)
...
                 Dto.TssState.UNINITIALIZED, _ => TssState.Uninitialized,
                 _ => throw new NotImplementedException($"Tss state: {state} is not implemented.")
             );
         }
+
+        private static ClientState MapClientState(Dto.ClientState state)
+        {
+            return state.Match(
+                Dto.ClientState.REGISTERED, _ => ClientState.Registered,
+                Dto.ClientState.DEREGISTERED, _ => ClientState.Deregistered
```
KaliCZ, line 115: What the hell is deregistered? 
Like Enabled and disabled? 
Or existing and deleted?

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -91,18 +96,18 @@ public Task<ResponseResult<Transaction>> StartTransactionAsync(AccessToken token
...
-                endpoint: $"tss/{tssId}/tx/{transactionId}",
+                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision=1",
                 request: RequestCreator.CreateTransaction(clientId),
                 successFunc: response => ModelMapper.MapTransaction(response),
                 token: token
             );
         }
 
-        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId, string lastRevision)
+        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId, string revision)
```
KaliCZ, line 106: 1. Shouldn't revision be an integer?
2. If we only allow Start and Finish, won't Finish always be called with `revision = "2"` ? Just like you did in tests? If yes, hardcode it here as well. If not, then I would add an XML comment and explain what it should be.

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
```diff
@@ -56,21 +61,21 @@ public Task<ResponseResult<Tss>> GetTssAsync(AccessToken token, Guid tssId)
             );
         }
 
-        public Task<ResponseResult<Tss>> CreateTssAsync(AccessToken token, TssState state, string description = null)
+        public Task<ResponseResult<CreateTss>> CreateTssAsync(AccessToken token)
         {
-            return Client.ProcessRequestAsync<Dto.CreateTssRequest, Dto.TssResponse, Tss>(
+            return Client.ProcessRequestAsync<object, Dto.CreateTssResponse, CreateTss>(
                 method: HttpMethod.Put,
                 endpoint: $"tss/{Guid.NewGuid()}",
```
KaliCZ, line 68: Didn't we agree that we will try to create TSS with the same ID as the original ones? Which means we need the ability to provide tssID, rather than getting a random GUID. cc @marektresnak as you're the one who requested him to remove it 😄

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
KaliCZ: But we want it for the TSS ID (as per the other comment) so why not allow it for the client too.

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
KaliCZ: I get that in theory, there can be updates. But if our library doesn't even support them, then I wouldn't care about it.

src/Germany/Mews.Fiscalizations.Germany/ModelMapper.cs
abdallahbeshi: Based on my understanding, they're more like Enabled and disabled.

Registered means the client can be used
Deregistered means the client cannot be used

The initial state when creating a client is Registered.

src/Germany/Mews.Fiscalizations.Germany/FiskalyClient.cs
abdallahbeshi: Yea that's what we said, but after looking at the usages in Mews, i saw no benefit from doing that.

Having the same TSS ID in both API versions doesn't bring any benifit.

My idea is to create a new TSS whenever there is an update on an enabled integration. (which makes sense, as you can update an integration with new API/Secret keys).

KaliCZ: Is there a benefit in creating a new TSS every time? It will make the dashboard look horrible.
There is a hidden benefit in having the TSS with the same ID. The simplicity. You have 1 Tss. And it's one ID. forever has been.

abdallahbeshi: I think it does make sense to create a new TSS each time you enable the integration.

This can be improved by:
1. Check if there is an existing TSS (if yes, use it).
2. Check if the API keys were changed, if yes (create a new TSS).

marektresnak: I didn't want to remove the parameter. What I wanted is to never generate a random guid in the library and make the parameter required instead 😄

abdallahbeshi:
Closing as we decided to go with another solution.

PR #77, opened 2021-09-23
Introduced a new folder structure for each API version

abdallahbeshi:
## Description

Introduced a new folder structure for each API version.

image

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.


src/Germany/Mews.Fiscalizations.Germany.Tests/TestFixture.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Mews.Fiscalizations.Germany.Model;
+﻿using Mews.Fiscalizations.Germany.V1;
+using Mews.Fiscalizations.Germany.V1.Model;
```
KaliCZ, line 2: This is weird. Why do we now need more usings? :D

src/Germany/Mews.Fiscalizations.Germany.Tests/TestFixture.cs
```diff
@@ -1,4 +1,5 @@
...
 using System;
 
 namespace Mews.Fiscalizations.Germany.Tests
```
KaliCZ, line 5: We could also copy paste the tests, so that tests have a v1 too.

abdallahbeshi: I would keep tests for the latest version only.

Having tests for all versions means we have to add more pipelines and maintain all of them.

src/Germany/Mews.Fiscalizations.Germany.Tests/TestFixture.cs
abdallahbeshi: `FiskalyClient` is no longer in Mews.Fiscalizations.Germany
it is now Mews.Fiscalizations.Germany.V1 so we must add the using.

the other using was there before.

KaliCZ (approved):

PR #78, opened 2021-09-23
Upgraded the library to use the new Fiskaly certified API v2

abdallahbeshi:
## Description

Upgraded the library to use the new Fiskaly certified API v2.

Fixes #RND-77090

## Type of change
- [ ] Bug fix.
- [x] Feature.
- [ ] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
+++ b/src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
@@ -103,11 +103,11 @@ namespace Mews.Fiscalizations.Germany.V2
             );
         }
 
-        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId, int revision = 2)
+        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId)
         {
             return Client.ProcessRequestAsync<Dto.FinishTransactionRequest, Dto.TransactionResponse, Transaction>(
                 method: HttpMethod.Put,
-                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision={revision}",
+                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision=2",
                 request: RequestCreator.FinishTransaction(clientId, bill),
                 successFunc: response => ModelMapper.MapTransaction(response),
                 token: token
--- a/src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
+++ b/src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
@@ -1,0 +1,159 @@
+﻿using FuncSharp;
+using Mews.Fiscalizations.Germany.V2.Model;
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Mews.Fiscalizations.Germany.V2
+{
+    public sealed class FiskalyClient
+    {
+        public ApiKey ApiKey { get; }
+        public ApiSecret ApiSecret { get; }
+
+        private Client Client { get; }
+
+        public FiskalyClient(ApiKey apiKey, ApiSecret apiSecret)
+        {
+            ApiKey = apiKey;
+            ApiSecret = apiSecret;
+            Client = new Client();
+        }
+
+        public Task<ResponseResult<Model.Client>> GetClientAsync(AccessToken token, Guid clientId, Guid tssId)
+        {
+            return Client.GetResponseAsync<Dto.ClientResponse, Model.Client>(
+                endpoint: $"tss/{tssId}/client/{clientId}",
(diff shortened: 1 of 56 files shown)


KaliCZ (approved):

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
```diff
@@ -91,18 +96,18 @@ public Task<ResponseResult<Transaction>> StartTransactionAsync(AccessToken token
...
             );
         }
 
-        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId, string lastRevision)
+        public Task<ResponseResult<Transaction>> FinishTransactionAsync(AccessToken token, Guid clientId, Guid tssId, Bill bill, Guid transactionId)
         {
             return Client.ProcessRequestAsync<Dto.FinishTransactionRequest, Dto.TransactionResponse, Transaction>(
                 method: HttpMethod.Put,
-                endpoint: $"tss/{tssId}/tx/{transactionId}?last_revision={lastRevision}",
+                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision=2",
```
marektresnak, line 110: This definitely shouldn't be hard-coded. The library should be universally usable and allow updates between transaction start and finish.

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
```diff
@@ -91,18 +96,18 @@ public Task<ResponseResult<Transaction>> StartTransactionAsync(AccessToken token
         {
             return Client.ProcessRequestAsync<Dto.TransactionRequest, Dto.TransactionResponse, Transaction>(
                 method: HttpMethod.Put,
-                endpoint: $"tss/{tssId}/tx/{transactionId}",
+                endpoint: $"tss/{tssId}/tx/{transactionId}?tx_revision=1",
```
marektresnak, line 99: The user of the client should be able to control the revision. So I would either:
1. Allow specifying the revision through a param.
2. Include the current revision in the response. Saying that a transaction always starts with revision 1 is a good solution, so I would probably do this.

src/Germany/Mews.Fiscalizations.Germany/V2/Model/ClientState.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Mews.Fiscalizations.Germany.V2.Model
+{
+    public enum ClientState
+    {
+        Enabled,
+        Disabled
```
marektresnak, line 6: Is there a reason to use different naming than the API? Considering the users of the lib will need to read documentation to understand the workflow, wouldn't it be easier to have the naming consistent with the API and official docs?

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
abdallahbeshi: That is right if we call the endpoint UpdateTransaction

But we call it FinishTransaction and actually everything is hard coded to finish the transaction.

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
abdallahbeshi: We can do it in a separate PR, but i mentioned this in the Readme and it is also mentioned in Fiskaly documentation.

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
marektresnak: But that is pretty specific to our use-case, right? The library should be universally usable.

src/Germany/Mews.Fiscalizations.Germany/V2/Model/ClientState.cs
abdallahbeshi: This model is what will be used on the client side (Mews in our case), In the DTO, they are called just like in Fiskaly documentation.

src/Germany/Mews.Fiscalizations.Germany/V2/FiskalyClient.cs
abdallahbeshi: Yes i completely agree, but changing this now won't help anyone, it just makes it confusing.

If we want to do it properly, we change rename this to UpdateTransactionAsync  and make it accept the transaction state and revision..etc but if we decide to do this, its definitely a separate PR.

src/Germany/Mews.Fiscalizations.Germany/V2/Model/ClientState.cs
marektresnak: I guess this is a general point on how we want to work with naming in the libraries. But as long as the original names are reasonable, I'd keep using those. We know that different naming in the lib vs the service causes confusion in some cases.

marektresnak (approved):

PR #97, opened 2021-11-18
Introduce soap action error handling, part 1

oleg-sokolansky:
## Description

This this part 1 of PR series which suppose to add handling of soap failures.

In scope: introduce result co-product with minimum changes.

Fixes [RND-80626](https://mews.myjetbrains.com/youtrack/issue/RND-80626)

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Spain/Mews.Fiscalizations.Spain.Tests/Basics.cs
+++ b/src/Spain/Mews.Fiscalizations.Spain.Tests/Basics.cs
@@ -70,9 +70,11 @@ namespace Mews.Fiscalizations.Spain.Tests.IssuedInvoices
 
             var response = await client.SubmitSimplifiedInvoiceAsync(model);
 
-            var responseErrorMessages = response.Invoices.Select(i => i.ErrorMessage).Flatten();
+            Assert.IsTrue(response.IsSuccess, $"Success response expected, but error received: {response.ErrorResult.Error}");
+
+            var responseErrorMessages = response.SuccessResult.Invoices.Select(i => i.ErrorMessage).Flatten();
             var errorMessage = String.Join(System.Environment.NewLine, responseErrorMessages);
-            Assert.AreEqual(response.Result, RegisterResult.Correct, errorMessage);
+            Assert.AreEqual(response.SuccessResult.Result, RegisterResult.Correct, errorMessage);
 
             return invoice;
         }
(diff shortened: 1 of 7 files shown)


src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
```diff
@@ -28,22 +30,33 @@ internal SoapClient(Uri endpointUri, X509Certificate certificate, TimeSpan httpT
...
+                return new ResponseResult<TOut>(successResult: deserializedMessage);
+            }
+            catch (InvalidOperationException)
+            {
+                var soapFault = XmlSerializer.Deserialize<SoapFaultResponse>(response);
+                return new ResponseResult<TOut>(errorResult: new ErrorResult(
+                    code: soapFault.Body.Fault.Faultcode,
+                    message: soapFault.Body.Fault.Faultstring
+                ));
+            }
```
abdallahbeshi, line 59: Unfortunately, the SII returns a success status code 200 when there is an issue with the XML (SOAPFault)

The same thing for validation errors, they are returned as a part of the invoice response.

abdallahbeshi (approved):

src/Spain/Mews.Fiscalizations.Spain/Nif/NifValidator.cs
```diff
@@ -17,11 +18,14 @@ public NifValidator(X509Certificate certificate, TimeSpan httpTimeout)
...
 
-        public async Task<Response> CheckNif(Request model)
+        public async Task<ResponseResult<Response>> CheckNif(Request model)
         {
             var request = Convert(model);
             var response = await SoapClient.SendAsync<Entrada, Salida>(request);
-            return Convert(request, response);
+            return response.IsSuccess.Match(
+                t => new ResponseResult<Response>(successResult: Convert(request, response.SuccessResult)),
+                f => new ResponseResult<Response>(errorResult: response.ErrorResult)
```
KaliCZ, line 27: I would prefer static methods over invocations with named optional parameters. I'd find that nicer.

But more importantly, ResponseResult now allows to be created with both a success result and an error result at the same time. That should not be possible.

src/Spain/Mews.Fiscalizations.Spain.Tests/Basics.cs
```diff
@@ -70,9 +70,9 @@ private async Task<SimplifiedInvoice> SuccessfullyPostInvoice(Client client)
 
             var response = await client.SubmitSimplifiedInvoiceAsync(model);
 
-            var responseErrorMessages = response.Invoices.Select(i => i.ErrorMessage).Flatten();
+            var responseErrorMessages = response.SuccessResult.Invoices.Select(i => i.ErrorMessage).Flatten();
```
KaliCZ, line 73: So then this is a null pointer if it fails. How did we improve the situation?

KaliCZ:
Will be fixed in followup PRs.

src/Spain/Mews.Fiscalizations.Spain.Tests/Basics.cs
oleg-sokolansky: There is used to be an assert which was checking whether response result is successful. However feedback was that it make test more confusing. Would you prefer to have it back?

KaliCZ: No, I prefer to have type safety, rather than contract check. So 2 public stacit methods - 1 success and 1 error and a private constructor. 
That way, even the user understands the usage without a need for any check.

PR #99, opened 2021-11-19
 Introduce soap action error handling, part 2.1

oleg-sokolansky:
## Description

This this part 2.1 of PR series which suppose to add handling of soap failures.

In scope: 
- add temporary test which should cover invalid response from sii. 
- response co-product creation via static helpers

Fixes [RND-80626](https://mews.myjetbrains.com/youtrack/issue/RND-80626)

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.


src/Spain/Mews.Fiscalizations.Spain/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Mews.Fiscalizations.Spain.Tests")]
```
abdallahbeshi, line 3: pls remove with the tests once u know it works.

src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
```diff
@@ -47,15 +47,19 @@ internal SoapClient(Uri endpointUri, X509Certificate certificate, TimeSpan httpT
...
-                    message: soapFault.Body.Fault.Faultstring
-                ));
+                try
+                {
+                    return ResponseResult.Error<TOut>(soapFault.Body.Fault.Code, soapFault.Body.Fault.Message);
+                }
+                catch (Exception e)
+                {
+                    throw new InvalidOperationException($"Fail to parse response {response}", e);
+                }
```
abdallahbeshi, line 62: IMO this is not needed, because if deserializing the response to SoapFaultResponse was successful, then we know we have these values as the server-side will always return them.

oleg-sokolansky: I put this just to help future investigations. If there is a fault in xml deserialization there is not way to get the message. I believe there is a decent chance of having an issue since we do not really consume code-generated service based on wsdl, but created all the contracts manually.

oleg-sokolansky: I would like to suggest to keep this at least at first time and remove afterwards.

abdallahbeshi (dismissed):

src/Spain/Mews.Fiscalizations.Spain/AssemblyInfo.cs
abdallahbeshi: @KaliCZ Pls ping @oleg-sokolansky before merging so he can remove the tests

src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
abdallahbeshi: @KaliCZ I really still think this doesn't make any sense and it won't help

if `var soapFault = XmlSerializer.Deserialize<SoapFaultResponse>(response);` succeeds, we must get a valid SOAP fault exeption https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.soapfault?view=netframework-4.8

KaliCZ: I don't understand what this try-catch is acually catching. It seems to me that the only thing it can catch is a null pointer in the `SoapFaultResponse` body.

In that case, I wouldn't really handle the null with a try-catch. There's 2 options here:
- It can never be null. In that case this is useless code. 
- It actually can be null. Then we should handle it properly and not with a general try-catch.

So I would get rid of it. And if it happens that nulls are happening, we can handle them.

src/Spain/Mews.Fiscalizations.Spain/Model/Response/ResponseResult.cs
```diff
@@ -2,6 +2,15 @@
 
 namespace Mews.Fiscalizations.Spain.Model.Response
 {
+    public static class ResponseResult
+    {
+        public static ResponseResult<T> Success<T>(T result) where T : class => new ResponseResult<T>(successResult: result);
+
+        public static ResponseResult<T> Error<T>(ErrorResult error) where T : class => new ResponseResult<T>(errorResult: error);
+
+        public static ResponseResult<T> Error<T>(string code, string message) where T : class => new ResponseResult<T>(errorResult: new ErrorResult(code, message));
```
KaliCZ, line 11: I kinda like this. But the lines are too long and the where doesn't help this. What do you think @marektresnak?

KaliCZ (approved):
Overall good. 
Let's remove the tests and the try-catch and it's good to go.

src/Spain/Mews.Fiscalizations.Spain/Model/Response/ResponseResult.cs
marektresnak: I do like it as well. Maybe we could just split to two lines? The long lines are difficult to read on some places (like here, in Github).

src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
```diff
@@ -43,19 +44,25 @@ internal SoapClient(Uri endpointUri, X509Certificate certificate, TimeSpan httpT
...
                 var deserializedMessage = XmlSerializer.Deserialize<TOut>(soapResponseBody.OuterXml);
-                return new ResponseResult<TOut>(successResult: deserializedMessage);
-            }
-            catch (InvalidOperationException)
+                return ResponseResult.Success(deserializedMessage);
+            });
+            return succesResult.Match(s => s, _ => DeserializeSoapError<TOut>(response));
+        }
+
+        private static ResponseResult<TOut> DeserializeSoapError<TOut>(string response) where TOut : class, new()
```
abdallahbeshi, line 56: Move `where TOut : class, new()` to a new line

src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
```diff
@@ -43,19 +44,25 @@ internal SoapClient(Uri endpointUri, X509Certificate certificate, TimeSpan httpT
...
                 var soapFault = XmlSerializer.Deserialize<SoapFaultResponse>(response);
-                return new ResponseResult<TOut>(errorResult: new ErrorResult(
-                    code: soapFault.Body.Fault.Faultcode,
-                    message: soapFault.Body.Fault.Faultstring
-                ));
+                return ResponseResult.Error<TOut>(soapFault.Body.Fault.Code, soapFault.Body.Fault.Message);
+            }
+            catch (Exception e)
+            {
+                throw new InvalidOperationException($"Fail to parse response {response}", e);
```
abdallahbeshi, line 65:
```suggestion
                throw new InvalidOperationException($"Failed to parse response.", e);
```

abdallahbeshi (approved):

src/Spain/Mews.Fiscalizations.Spain/Communication/SoapClient.cs
```diff
@@ -62,7 +62,7 @@ private static ResponseResult<TOut> DeserializeSoapError<TOut>(string response)
             }
             catch (Exception e)
             {
-                throw new InvalidOperationException($"Fail to parse response {response}", e);
+                throw new InvalidOperationException($"Failed to parse response.", e);
```
KaliCZ, line 66: Hold on, why did we drop the response? That was the only justification of this whole exception catching. If we don't expose the response, the whole try catch is useless.

KaliCZ (approved):

TashaBubo:
@KaliCZ `tested`?

KaliCZ:
@TashaBubo Can you actually exclude PRs from this repository from your board completely?

PR #109, opened 2022-01-18
Introduced Uniwix error handling

abdallahbeshi:
## Description

Instead of throwing exceptions in the library, an ITry with the expected response and error result will be returned to the client-side.

Fixes #issue

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.


src/Italy/Mews.Fiscalizations.Italy/Uniwix/Communication/UniwixClient.cs
```diff
@@ -79,52 +74,56 @@ public Task<UniwixUser> CreateUserAsync(CreateUserParameters createUserParameter
...
         }
 
-        private async Task<TResult> ExecuteRequestAsync<TResult>(string url, HttpMethod httpMethod, HttpContent content)
+        private Task<ITry<TResult, ErrorResult>> ExecuteRequestAsync<TResult>(string url, HttpMethod httpMethod, HttpContent content)
         {
-            var response = await ExecuteRequestAsync(url, httpMethod, content, async httpResponse =>
+            return ExecuteRequestAsync(url, httpMethod, content, httpResponse =>
             {
-                var json = await httpResponse.Content.ReadAsStringAsync();
+                var json = httpResponse.Content.ReadAsStringAsync().Result;
```
oleg-sokolansky, line 98: I would like to double check whether it is intentional. When Result of incomplete task is accessed I believe it would make this operation synchronous. Could you please consider keeping it async and await on line 142?

oleg-sokolansky (approved):

src/Italy/Mews.Fiscalizations.Italy.Tests/UniwixClientTests.cs
```diff
@@ -35,23 +34,32 @@ public async Task SendInvoiceSucceeds()
                 Body = new[] { GetInvoiceBody(invoiceNumber) }
             });
 
-            var fileId = result.FileId;
-            Assert.IsNotEmpty(result.FileId);
-            Assert.IsNotEmpty(result.Message);
-
-            Thread.Sleep(500);
```
KaliCZ, line 42: Isn't it gonna fail without the thread sleep?

src/Italy/Mews.Fiscalizations.Italy/Uniwix/Communication/UniwixClient.cs
```diff
@@ -79,52 +74,56 @@ public Task<UniwixUser> CreateUserAsync(CreateUserParameters createUserParameter
...
         }
 
-        public Task<bool> VerifyCredentialsAsync()
+        public async Task<ITry<bool, ErrorResult>> VerifyCredentialsAsync()
         {
-            var url = $"{UniwixBaseUrl}/Info";
-            return ExecuteRequestAsync(url, HttpMethod.Get, content: null, responseProcessor: r => r.IsSuccessStatusCode);
+            // This endpoint returns the account information, so if the ITry is success it means the credentials are valid.
+            var result = await GetAsync<object>($"{UniwixBaseUrl}/Info");
+            return result.Map(r => true);
```
KaliCZ, line 81: Shouldn't this still be like this?
```suggestion
            return result.Map(r => r.IsSuccessStatusCode);
```

src/Italy/Mews.Fiscalizations.Italy/Uniwix/Communication/UniwixClient.cs
```diff
@@ -41,32 +43,25 @@ public async Task<SendInvoiceResult> SendInvoiceAsync(ElectronicInvoice invoice)
...
+        public async Task<ITry<InvoiceState, ErrorResult>> GetInvoiceStateAsync(string fileId)
         {
             var url = $"{UniwixBaseUrl}/Invoices/{fileId}";
             var result = await GetAsync<List<InvoiceStateResult>>(url);
 
-            if (result.Count == 0)
-            {
-                throw new InvalidOperationException($"Invoice {fileId} not found.");
-            }
+            var validatedResult = result.FlatMap(r => r.ToTry(a => a.NonEmpty(), _ => ErrorResult.Create($"Invoice {fileId} not found.", ErrorType.InvoiceNotFound)));
```
KaliCZ, line 55: Isn't there a Where method for this? So you wouldn't have to use the ToTry inside?

KaliCZ (approved):

src/Italy/Mews.Fiscalizations.Italy/Uniwix/Communication/UniwixClient.cs
abdallahbeshi: This is handled in `ExecuteRequestAsync`, if anything is wrong, it will be in the Error part, otherwise we know its always true (if we got a valid response, then we know the credentials are valid).

If we ever needed this response to return the account information, all we have to do is create the model and map the response. but for now its good enough to say if we have a valid response, then the result is true (valid credentials).

KaliCZ: But you only changed the way exceptions are handled. So if it was possible before that it would be false here, then it's still possible.

KaliCZ (approved):

PR #139, opened 2022-07-04
Introduced TicketBai send invoice response model

abdallahbeshi:
## Description

Introduced TicketBai send invoice response model.

## Type of change
- [ ] Bug fix.
- [x Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
+++ b/src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
@@ -1,0 +1,31 @@
+﻿using FuncSharp;
+using Mews.Fiscalizations.TicketBai.Model;
+using System;
+using System.Linq;
+
+namespace Mews.Fiscalizations.TicketBai
+{
+    internal static class DtoToModelConverter
+    {
+        public static SendInvoiceResponse Convert(Dto.TicketBaiResponse response, string qrCodeUri, string xmlRequestContent)
+        {
+            var result = response.Salida;
+            return new SendInvoiceResponse(
+                xmlRequestContent: xmlRequestContent,
+                qrCodeUri: qrCodeUri,
+                tbaiIdentifier: result.IdentificadorTBAI,
+                received: DateTime.Parse(result.FechaRecepcion),
+                state: result.Estado,
+                description: result.Descripcion,
+                stateExplanation: result.Azalpena,
+                csv: result.CSV,
+                validationResults: result.ResultadosValidacion?.Select(v => Convert(v))
+            );
+        }
+
+        private static SendInvoiceValidationResult Convert(Dto.ResultadosValidacion validation)
+        {
+            return new SendInvoiceValidationResult(validation.Codigo, validation.Descripcion, validation.Azalpena);
+        }
+    }
+}
--- a/src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceResponse.cs
+++ b/src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceResponse.cs
@@ -1,0 +1,49 @@
+﻿using FuncSharp;
+using System;
+using System.Collections.Generic;
+
+namespace Mews.Fiscalizations.TicketBai.Model
+{
+    public sealed class SendInvoiceResponse
+    {
+        public SendInvoiceResponse(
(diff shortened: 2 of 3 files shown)


src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceResponse.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public string TBAIIdentifier { get; }
+
+        public DateTime Received { get; }
+
+        public string State { get; } // TODO: enum?
+
+        public string Description { get; }
+
+        public string StateExplanation { get; }
```
KaliCZ, line 43: Maybe also an enum?

src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceValidationResult.cs
```diff
@@ -0,0 +1,18 @@
...
+            ErrorCode = errorCode;
+            Description = description;
+            Explanation = explanation;
+        }
+
+        public string ErrorCode { get; } // TODO: enum?
+
+        public string Description { get; }
+
+        public string Explanation { get; }
```
KaliCZ, line 16: ALso maybe an enum?

src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
```diff
@@ -0,0 +1,31 @@
+﻿using FuncSharp;
+using Mews.Fiscalizations.TicketBai.Model;
+using System;
+using System.Linq;
+
+namespace Mews.Fiscalizations.TicketBai
+{
+    internal static class DtoToModelConverter
+    {
+        public static SendInvoiceResponse Convert(Dto.TicketBaiResponse response, string qrCodeUri, string xmlRequestContent)
```
KaliCZ, line 10: So we're returning the requestContent into the call site so that the usages can log what was sent right? 
In that case it would also make sense to return the actual request?

Also, where are we getting the QR code from, if not the server response? 😄

KaliCZ (approved):

src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
abdallahbeshi: We will be creating the QR code based on some rules defined in the documentation

its a combination of stuff we define/calculate and stuff returned in the response.

src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceResponse.cs
abdallahbeshi: This will contain a text explaining the state.

src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceValidationResult.cs
abdallahbeshi: same here

src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
KaliCZ: But this says qrCodeUri - so it must be somewhere hosted on some URL. So we get sometihng from the response and then upload something somewhere in a second request and get the qr code uri?

src/TicketBai/Mews.Fiscalizations.TicketBai/Model/SendInvoiceResponse.cs
KaliCZ: Sure, but there's like a predefined list of answers. So it theoretically could be like an Enum. Hey, we'll see if there's something later on.

src/TicketBai/Mews.Fiscalizations.TicketBai/DtoToModelConverter.cs
KaliCZ: Anyway, we'll go through the model again later, no need to worry about it now :)

PR #147, opened 2022-08-03
Basque fiscalization cleanup

abdallahbeshi:
## Description

Fixed comments from previous PRs.

Fixes #issue

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Basque/Mews.Fiscalizations.Basque.Tests/SendInvoiceTests.cs
+++ b/src/Basque/Mews.Fiscalizations.Basque.Tests/SendInvoiceTests.cs
@@ -13,7 +13,6 @@ namespace Mews.Fiscalizations.Basque.Tests
             var request = CreateInvoiceRequest();
             var response = await Client.SendInvoiceAsync(request);
 
-            // TODO: Assert the Qr code uri content, make sure it contains all the required params/data.
             Assert.IsEmpty(response.ValidationResults.Flatten());
             Assert.IsNotEmpty(response.QrCodeUri);
             Assert.IsNotEmpty(response.TBAIIdentifier);
--- a/src/Basque/Mews.Fiscalizations.Basque.Tests/SendInvoiceTests.cs
+++ b/src/Basque/Mews.Fiscalizations.Basque.Tests/SendInvoiceTests.cs
@@ -1,32 +1,12 @@
 ﻿using FuncSharp;
-using Mews.Fiscalizations.Core.Model;
-using Mews.Fiscalizations.Basque.Model;
 using NUnit.Framework;
-using System;
-using System.Linq;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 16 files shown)


src/Basque/Mews.Fiscalizations.Basque.Tests/SendInvoiceTests.cs
```diff
@@ -1,148 +1,22 @@
...
-using System;
-using System.Linq;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 
 namespace Mews.Fiscalizations.Basque.Tests
 {
     [TestFixture]
-    public class Tests
+    public class SendInvoiceTests : TestFixture
```
KaliCZ, line 8: I'd rather not use it via inheritance, but a standard static call.

src/Basque/Mews.Fiscalizations.Basque/Mews.Fiscalizations.Basque.csproj
```diff
@@ -10,7 +10,7 @@
     <RepositoryUrl>https://github.com/MewsSystems/fiscalizations</RepositoryUrl>
     <Icon>https://raw.githubusercontent.com/msigut/eet/master/receipt.png</Icon>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageVersion>2.0.0-alpha</PackageVersion>
+    <PackageVersion>1.0.0</PackageVersion>
```
KaliCZ, line 13: So after alpha, we can start from 1? Are you sure?

src/Basque/Mews.Fiscalizations.Basque.Tests/TestFixture.cs
```diff
@@ -0,0 +1,131 @@
...
+                taxRatePercentage: Percentage.Create(vat).Success.Get(),
+                taxBaseAmount: Amount.Create(baseValue).Success.Get(),
+                taxAmount: Amount.Create(Math.Round(baseValue * vat / 100, 2)).Success.Get()
+            );
+        }
+
+        private static string RandomString(int length)
+        {
+            var chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+            return new string(Enumerable.Repeat(chars, length).Select(s => s[Random.Next(s.Length)]).ToArray());
```
KaliCZ, line 128: I'll let you read this again, so you can thing about what you want it to do 😁

KaliCZ: But like generally, why do we want a random string anyway? Either we can hardcode whatever we want, or we need it to be unique in which case we should use Guid.

KaliCZ (approved):

KaliCZ (approved):

PR #154, opened 2022-08-25
Fixed the Hungarian tax id regex

abdallahbeshi:
## Description

As per the technical documentation, the HU tax id must only consist of 8 numbers
image

So we extended the tax id validation to validate based on the client usage, either by allowing the country code prefix in the tax id or without.

## Type of change
- [x] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.


danielchorous:
@abdallahbeshi we should also fix BE crashing in case of an unsuccessful request. Let's have a call about this?

abdallahbeshi:
> @abdallahbeshi we should also fix BE crashing in case of an unsuccessful request. Let's have a call about this?

We actually want it to throw an error if the library is not handling some response so we get notified and fix it.

KaliCZ:
@abdallahbeshi Correct. But now we know that there is a possibility for the server to return this type of response. So we should handle this type of response.

abdallahbeshi:
> 

Yea we knew that before, the server returned an invalid request because we sent an invalid XML.

I can make it handle the "invalid request" response we get from the server, but i don't think that's a priority now, so pls create a YT

KaliCZ:
What do you mean invalid XML? the XML schema didn't change right? This value was valid before and now the value is invalid. But the web service is still the same with the same XML schema.

KaliCZ:
Or like it was crashing the whole time.

abdallahbeshi:
> What do you mean invalid XML? the XML schema didn't change right? This value was valid before and now the value is invalid. But the web service is still the same with the same XML schema.

The XML schema didn't change, it's just that we got lucky that no one ever provided an invalid HU tax ID :D

src/Core/Mews.Fiscalizations.Core/Model/Countries/Countries.cs
```diff
@@ -110,7 +110,7 @@ static Countries()
                 HolySee = NonEuropean("VA"),
                 Honduras = NonEuropean("HN"),
                 HongKong = NonEuropean("HK"),
-                Hungary = European("HU", "^(HU)?[0-9]{8}$"),
+                Hungary = European("HU", "^[0-9]{8}$"),
```
KaliCZ, line 113: The HU has to be allowed. So we can create a second regex and make Hungary use the second one. But generally, it's OK to provide the country code in TaxID.

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/Receiver.cs
```diff
@@ -15,7 +15,7 @@ public Receiver(Company company)
         {
         }
 
-        public static ITry<Receiver, Error> LocalCompany(TaxpayerIdentificationNumber taxpayerId, Name name, SimpleAddress address)
+        public static ITry<Receiver, Error> LocalCompany(string taxpayerId, Name name, SimpleAddress address)
```
KaliCZ, line 18: Does this simplify the usage?

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/LocalCompany.cs
```diff
@@ -19,12 +18,9 @@ private LocalCompany(TaxpayerIdentificationNumber taxpayerId, Name name, SimpleA
 
         public SimpleAddress Address { get; }
 
-        public static ITry<LocalCompany, Error> Create(TaxpayerIdentificationNumber taxpayerId, Name name, SimpleAddress address)
+        public static ITry<LocalCompany, Error> Create(string taxpayerId, Name name, SimpleAddress address)
```
KaliCZ, line 21: Same question.

src/Core/Mews.Fiscalizations.Core/Model/Countries/EuropeanUnionCountry.cs
```diff
@@ -5,15 +5,18 @@ namespace Mews.Fiscalizations.Core.Model
 {
     public class EuropeanUnionCountry
     {
-        internal EuropeanUnionCountry(string alpha2Code, Regex taxpayerNumberPattern)
+        internal EuropeanUnionCountry(string alpha2Code, Regex regexWithCountryCodePrefix, Regex regexWithoutCountryCodePrefix)
         {
             Alpha2Code = alpha2Code;
-            TaxpayerNumberPattern = taxpayerNumberPattern;
+            RegexWithCountryCodePrefix = regexWithCountryCodePrefix;
+            RegexWithoutCountryCodePrefix = regexWithoutCountryCodePrefix;
```
KaliCZ, line 12: Right, but we need to name it better. This class is a country. It's not obvious, what the regex is for.
So maybe `taxpayerNumberPattern` and `taxpayerNumberPatternWithoutCountryCodePrefix`?

KaliCZ (approved):

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/LocalCompany.cs
```diff
@@ -19,12 +18,9 @@ private LocalCompany(TaxpayerIdentificationNumber taxpayerId, Name name, SimpleA
...
         {
-            var localTaxPayerId = taxpayerId.ToOption().Where(i => i.Country.Equals(Countries.Hungary));
-            return localTaxPayerId.ToTry(_ => new Error($"{nameof(TaxpayerIdentificationNumber)} must be a Hungarian taxpayer number.")).Map(
-                i => new LocalCompany(i, name, address)
-            );
+            return TaxpayerIdentificationNumber.Create(Countries.Hungary, taxpayerId).Map(n => new LocalCompany(n, name, address));
```
KaliCZ, line 23: Shouldn't this actually say AllowsPrefix: false?

src/Core/Mews.Fiscalizations.Core/Model/Legal/EuropeanUnionTaxpayerIdentificationNumber.cs
```diff
@@ -14,11 +14,20 @@ private EuropeanUnionTaxpayerIdentificationNumber(EuropeanUnionCountry country,
...
+            EuropeanUnionCountry country,
+            string taxpayerNumber,
+            bool isCountryCodePrefixAllowed = true)
         {
             return ObjectValidations.NotNull(country).FlatMap(c =>
             {
-                var validNumber = StringValidations.RegexMatch(taxpayerNumber, country.TaxpayerNumberPattern);
+                var validNumber = StringValidations.RegexMatch(
+                    value: taxpayerNumber,
+                    pattern: isCountryCodePrefixAllowed.Match(
```
KaliCZ, line 26: I'd just create a variable pattern and leave this call as a single-line

src/Hungary/Mews.Fiscalizations.Hungary/Models/Invoice/Receiver.cs
abdallahbeshi: Yea it should as you don't have to validate the tax id and then the local company
there is an example in the modified tests

KaliCZ (approved):

PR #166, opened 2022-10-14
Fixed a typo in Basque fiscalization region name

abdallahbeshi:
## Description

Fixed a typo in the Basque fiscalization region name

## Type of change
- [x] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Core/Mews.Fiscalizations.Core/Model/Legal/EuropeanUnionTaxpayerIdentificationNumber.cs
+++ b/src/Core/Mews.Fiscalizations.Core/Model/Legal/EuropeanUnionTaxpayerIdentificationNumber.cs
@@ -25,7 +25,7 @@ namespace Mews.Fiscalizations.Core.Model
                     t => country.TaxpayerNumberPattern,
                     f => country.TaxpayerNumberPatternWithoutCountryCodePrefix
                 );
-                return StringValidations.RegexMatch(taxpayerNumber, pattern).Map(n => new EuropeanUnionTaxpayerIdentificationNumber(c, n));
+                return StringValidations.RegexMatch(taxpayerNumber.Trim(), pattern).Map(n => new EuropeanUnionTaxpayerIdentificationNumber(c, n));
             });
         }
     }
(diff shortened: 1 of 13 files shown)


src/Core/Mews.Fiscalizations.Core/Model/Legal/EuropeanUnionTaxpayerIdentificationNumber.cs
```diff
@@ -25,7 +25,7 @@ private EuropeanUnionTaxpayerIdentificationNumber(EuropeanUnionCountry country,
                     t => country.TaxpayerNumberPattern,
                     f => country.TaxpayerNumberPatternWithoutCountryCodePrefix
                 );
-                return StringValidations.RegexMatch(taxpayerNumber, pattern).Map(n => new EuropeanUnionTaxpayerIdentificationNumber(c, n));
+                return StringValidations.RegexMatch(taxpayerNumber.Trim(), pattern).Map(n => new EuropeanUnionTaxpayerIdentificationNumber(c, n));
```
KaliCZ, line 28: I don't like the trimming. Is it necessary? We're validating stuff but this doesn't cause error, it just corrects it instead.

abdallahbeshi: I believe it should be completely ok to fill the tax id as CZ 1234... and not CZ1234.....

or its possible that someone made amistake and added an extra space at the end (which happens A LOT), so correcting these cases should be fine IMO.

this would still fail if you only insert whitespaces anyway.

KaliCZ: As discussed on call, it's inconsistent - we validate nulls but then trim. So we should either just ignore nulls and replace with empty string or validate both or something. 

We're moving the trim to the call site.

KaliCZ (dismissed):

KaliCZ (approved):

PR #172, opened 2022-10-24
Bumped the lib versions to use latest core.

abdallahbeshi:
## Description

Bumped the lib versions to use the latest core lib.

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Germany/Mews.Fiscalizations.Germany/V2/RequestCreator.cs
+++ b/src/Germany/Mews.Fiscalizations.Germany/V2/RequestCreator.cs
@@ -94,7 +94,7 @@ namespace Mews.Fiscalizations.Germany.V2
         {
             return new Dto.AmountsPerPaymentType
             {
-                Amount = (payment.Amount + 0.00m).ToString(CultureInfo.InvariantCulture),
+                Amount = ToString(payment.Amount),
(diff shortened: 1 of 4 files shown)


src/Germany/Mews.Fiscalizations.Germany/V2/RequestCreator.cs
```diff
@@ -152,5 +152,10 @@ private static Dto.ClientState SerializeClientState(ClientState state)
                 ClientState.Deregistered, _ => Dto.ClientState.DEREGISTERED
             );
         }
+
+        private static string ToString(decimal amount)
+        {
+            return String.Format(CultureInfo.InvariantCulture, "{0:F2}", amount);
```
abdallahbeshi, line 158: Sending an amount with more than 2-5 decimals would violate fiskaly scheme (validation). so we should round the decimal.

KaliCZ: So why not round to 5 then? Why 2?

KaliCZ (approved):

PR #181, opened 2022-11-28
Removed deprecated HttpWebRequest from Austrian fiscalization and replaced it with HttpClient 

abdallahbeshi:
## Description

Removed deprecated HttpWebRequest from Austrian fiscalization and replaced it with HttpClient.

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.


.github/workflows/publish-all.yml
```diff
@@ -25,10 +25,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
     
-    - name: Setup .NET 6
-      uses: actions/setup-dotnet@v2
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v3
       with:
-        dotnet-version: 6.0.x
+        dotnet-version: 3.1.x
```
abdallahbeshi, line 31: Reverting back to .net core 3.1.x as dotnet nuget publish --skip-duplicates doesn't work with .net 6 for some reason :D

src/Austria/Mews.Fiscalizations.Austria/ATrust/SimpleHttpClient.cs
```diff
@@ -1,59 +0,0 @@
...
-
-        public string GetJson(string operation)
-        {
-            var request = CreateGetRequest($"{EndpointUrl.Value}/{operation}");
-            return GetResponseBody(request);
-        }
-
-        private HttpWebRequest CreateRequest(string method, string url)
-        {
-            var webRequest = WebRequest.Create(url) as HttpWebRequest;
```
abdallahbeshi, line 30: Deprecated

src/Austria/Mews.Fiscalizations.Austria/Dto/Receipt.cs
```diff
@@ -141,7 +141,7 @@ private ChainValue ComputeChainValue()
 
         private byte[] Sha256(byte[] toBeHashed)
         {
-            return new SHA256Managed().ComputeHash(toBeHashed);
+            return SHA256.Create().ComputeHash(toBeHashed);
```
abdallahbeshi, line 144: deprecated

marioporcic (approved):

.github/workflows/publish-all.yml
marektresnak: Does it mean that we'll be stuck on 3.1? 😄 I tried googling very quickly and found issues like [this](https://github.com/microsoft/azure-pipelines-tasks/issues/13591). Would [the solution](https://github.com/microsoft/azure-pipelines-tasks/issues/13591#issuecomment-793933011) with `allowPackageConflicts` help in this case?

marektresnak (commented):
The AT library really shows its age, so this is a nice update 👏

.github/workflows/publish-all.yml
abdallahbeshi: _Does it mean that we'll be stuck on 3.1?_ Not really, I will need to figure out why `--skip-duplicates` doesn't work as it should.

I will revisit this sometime later, i have it on my todo list :) but for now, we can keep the publish pipelines to use .net core 3.1, that shouldn't be an issue for now.

marektresnak (approved):

PR #185, opened 2022-12-16
Fixed stripping diacritics from IT fiscalization DTOs

abdallahbeshi:
## Description

Fixed stripping diacritics from IT fiscalization DTOs.

## Type of change
- [x] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Italy/Mews.Fiscalizations.Italy/Dto/DtoUtils.cs
+++ b/src/Italy/Mews.Fiscalizations.Italy/Dto/DtoUtils.cs
@@ -23,9 +23,9 @@ namespace Mews.Fiscalizations.Italy.Dto
             return Regex.Replace(zip, "[^0-9]", "");
         }
 
-        public static string NonEmptyValueOrNull(this string str)
+        public static string NonEmptyValueOrNull(this string s)
         {
-            return String.IsNullOrEmpty(str) ? null : str;
+            return String.IsNullOrEmpty(s) ? null : s;
         }
 
         public static string StripDiacritics(this string s)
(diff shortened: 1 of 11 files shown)


tonyjoanes (approved):

src/Italy/Mews.Fiscalizations.Italy/Dto/DtoUtils.cs
```diff
@@ -26,33 +23,20 @@ public static string NormalizeZip(this string zip)
...
-            return String.IsNullOrEmpty(str) ? null : str;
-        }
-
-        private static char GetCharacterSubstitute(char character, char substituteCharacter, int highestValidCharacter)
-        {
-            return (int)character <= highestValidCharacter ? character : substituteCharacter;
-        }
+            if (string.IsNullOrWhiteSpace(s))
+            {
+                return s.NonEmptyValueOrNull();
```
KaliCZ, line 35: Why would we replace empty string with a null?

KaliCZ (commented):
So just to confirm, you're sure that we no longer care about the max ascii character?

abdallahbeshi:
> So just to confirm, you're sure that we no longer care about the max ascii character?

@KaliCZ The technical docs doesn't say anything about that

src/Italy/Mews.Fiscalizations.Italy/Dto/DtoUtils.cs
abdallahbeshi: Yea that looks fishy, i just left as is for now, but will have a ticket to check the tech documentation and see if its needed or not..

PR #198, opened 2023-08-07
Funcsharp without covariance + bumped packages

KaliCZ:
## Description

Bumps Funscharp version.

Fixes #issue

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Mews.Fiscalizations.All.props
+++ b/src/Mews.Fiscalizations.All.props
@@ -3,6 +3,6 @@
     <AddSyntheticProjectReferencesForSolutionDependencies>false</AddSyntheticProjectReferencesForSolutionDependencies>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="FuncSharp" Version="[6.0.0, 7)" />
+    <PackageReference Include="FuncSharp" Version="7.0.1" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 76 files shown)


jbrestan:
Hmm, this will leak the preview build of FuncSharp to all Mews.Fiscalization consumers. Do you want that? OTOH I don't know how else to do this than making this preview as well

src/Mews.Fiscalizations.All.props
```diff
@@ -3,6 +3,6 @@
     <AddSyntheticProjectReferencesForSolutionDependencies>false</AddSyntheticProjectReferencesForSolutionDependencies>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="FuncSharp" Version="[6.0.0, 7)" />
+    <PackageReference Include="FuncSharp" Version="7.0.1" />
```
jbrestan, line 6: Shouldn't it be this?
```suggestion
    <PackageReference Include="FuncSharp" Version="[7.0.0-preview.1]" />
```

KaliCZ: It should. I was reverting the packages, so I forgot about this. 
I did use the correct one in Globalization though 😄

jbrestan: That's where I copied it from 😅

jbrestan (approved):

PR #215, opened 2023-09-13
Fixed tbai signature signing

abdallahbeshi:
## Description

A temporary fix for signing ticketbai xml using XadES-BES

## Type of change
- [x] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.


inakiesteve (approved):
Only problem I can see here is the signing is coupled to the Basque Regions but if we ever need to apply these to other potential fiscalizations, it should be easy to create an abstraction like you mentioned yesterday.

PR #225, opened 2023-10-23
Introduced bizkaia header dto + support for gzip compression

inakiesteve:
## Description

I have included support for the Bizkaia header dto that should be included in the header of every request to Batuz.
I have also included support for gzipping and gunzipping strings in the library. 

Tests have been included to test:
- Json serialization of the BizkaiaData class.
- Compress and decompress work as expected. 

## Type of change
- [ ] Bug fix.
- [x ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.
- [ x] Tests included (please specify cases).


src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+
+                memoryStream.Position = 0;
+                using (var gZipStream = new GZipStream(memoryStream, CompressionMode.Decompress))
+                {
+                    gZipStream.Read(buffer, 0, buffer.Length);
+                }
+
+                return encoding.GetString(buffer);
+            }
+        }
```
abdallahbeshi, line 47: Same here, how about this

```
        public static string Decompress(byte[] bytes, Encoding encoding)
        {
            using var memoryStream = new MemoryStream(bytes);
            using var outputStream = new MemoryStream();
            using (var decompressStream = new GZipStream(memoryStream, CompressionMode.Decompress))
            {
                decompressStream.CopyTo(outputStream);
            }
            return encoding.GetString(outputStream.ToArray());
        }
```

src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+            memoryStream.Position = 0;
+
+            var compressedData = new byte[memoryStream.Length];
+            memoryStream.Read(compressedData, 0, compressedData.Length);
+
+            var gZipBuffer = new byte[compressedData.Length + 4];
+            Buffer.BlockCopy(compressedData, 0, gZipBuffer, 4, compressedData.Length);
+            Buffer.BlockCopy(BitConverter.GetBytes(buffer.Length), 0, gZipBuffer, 0, 4);
+            return Convert.ToBase64String(gZipBuffer);
+        }
```
abdallahbeshi, line 27: This code is really hard to read and understand, also i believe it contains lots of weird logic

How about:

```
        public static byte[] Compress(string content, Encoding encoding)
        {
            var bytes = encoding.GetBytes(content);
            using var memoryStream = new MemoryStream();
            using (var gzipStream = new GZipStream(memoryStream, CompressionMode.Compress))
            {
                gzipStream.Write(bytes);
            }
            return memoryStream.ToArray();
        }
```

src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
abdallahbeshi: And now since the codes are almost identical, maybe you can make it accept the CompressionMode as a param and have one method instead of 2? 😄 not sure if that's great but consider it

src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.IO.Compression;
+using System.Text;
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Mews.Fiscalizations.Core.Compression
+{
+    public static class GzipCompressionHelper
```
abdallahbeshi, line 7: How about `GZipCompressorHelper`

src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
abdallahbeshi: one more thing, how about u make this a string extension? so you would do "hello world".Compress...etc

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Mews.Fiscalizations.Core.Tests.Compression;
+
+[TestFixture]
+public class GzipCompressionTests
+{
+    private const string Message = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+
+    [Test]
+    public void GzipCompression_Succeeds()
```
abdallahbeshi, line 11: GzipCompression_Succeeds is not so clear, here you're compressing and decompressing

and it should be clear that you're compressing a string

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        Assert.DoesNotThrow(() =>
+        {
+            string compressedString = GzipCompressionHelper.Compress(
+                inputString: Message
+                );
```
abdallahbeshi, line 17: Can be a one-liner, or as you wrote it but fix the closing parenthesis indentation

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+            string decompressedString = GzipCompressionHelper.Decompress(
+                compressedString: compressedString,
+                encoding: System.Text.Encoding.ASCII
+                );
```
abdallahbeshi, line 22: Can be a one-liner, or as you wrote it but fix the closing parenthesis indentation

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaData.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Mews.Fiscalizations.Basque.Dto.Bizkaia.Header;
+
+public partial class BizkaiaData
```
abdallahbeshi, line 3: Isn't this supposed to be provided in the header?

If yes, then BizkaiaHeaderData makes more sense?

src/Basque/Mews.Fiscalizations.Basque.Tests/Bizkaia/Helpers/BizkaiaDataHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+                ap1 = "Family name 1",
+                ap2 = "Family name 2"
+            },
+            drs = new Drs
+            {
+                mode = "140/240",
+                ejer = "ejercicio"
+            }
+        };
+    }
```
abdallahbeshi, line 26: Why don't you just inline this in the test?

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,27 @@
...
+public class GzipCompressionTests
+{
+    private const string Message = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+
+    [Test]
+    public void GzipCompression_Succeeds()
+    {
+        Assert.DoesNotThrow(() =>
+        {
+            string compressedString = GzipCompressionHelper.Compress(
```
abdallahbeshi, line 15: I already commented this on multiple PRs many times, please use var when possible.

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,27 @@
...
+                inputString: Message
+                );
+
+            string decompressedString = GzipCompressionHelper.Decompress(
```
abdallahbeshi, line 19: var

src/Basque/Mews.Fiscalizations.Basque.Tests/Bizkaia/BizkaiaDataValidationTests.cs
```diff
@@ -0,0 +1,21 @@
...
+public class BizkaiaDataValidationTests
+{
+    [Test]
+    public void Create_BizkaiaData_JsonSerialization_Succeeds()
+    {
+        var bizkaiaData = BizkaiaDataHelper.CreateSampleBizkaiaData();
+
+        Assert.DoesNotThrow(() =>
+        {
+            string serializedData = JsonSerializer.Serialize(bizkaiaData);
```
abdallahbeshi, line 17: var

src/Basque/Mews.Fiscalizations.Basque.Tests/Bizkaia/BizkaiaDataValidationTests.cs
```diff
@@ -0,0 +1,21 @@
...
+            Assert.IsFalse(string.IsNullOrEmpty(serializedData));
```
abdallahbeshi, line 18: Would `JsonSerializer.Serialize` return null in this case?
Anyways, i would use `Assert.IsNotEmpty`

abdallahbeshi (changes requested):

src/Core/Mews.Fiscalizations.Core/Compression/GzipCompressionHelper.cs
inakiesteve: The difference is that the both methods takes and return different types. We could abstract parts of the code though, but I don't think it is really worth it.

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Mews.Fiscalizations.Basque.Dto.Bizkaia.Header;
+
+public class BizkaiaHeaderData
+{
+    [JsonPropertyName("con")]
+    public string With { get; set; }
```
abdallahbeshi, line 8: The value of this will be always LROE, no? if yes, then no need for the setter

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+    
+    [JsonPropertyName("apa")]
+    public string Section { get; set; }
+
+    [JsonPropertyName("inte")]
+    public Inte Inte { get; set; }
```
abdallahbeshi, line 14: Is it possible to translate this to Eng?

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    [JsonPropertyName("drs")]
+    public Drs Drs { get; set; }
```
abdallahbeshi, line 17: same here

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+public class Drs
+{
+    [JsonPropertyName("mode")]
+    public string Mode { get; set; }
```
abdallahbeshi, line 23: Isn't this the model type 140/240? if yes, then I'd go with an enum or if we're supporting only one of them, then just hardcode it

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    [JsonPropertyName("ejer")]
+    public string FiscalYear { get; set; }
```
abdallahbeshi, line 26: If its only the year, then this can be an int

and any idea what should we provide here exactly? if it's gonna be "current year", then maybe we can also hardcode it.

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+public class Inte
+{
+    [JsonPropertyName("nif")]
+    public string TaxPayerIdentificationNumber { get; set; }
```
abdallahbeshi, line 32:
```suggestion
    public string TaxpayerIdentificationNumber { get; set; }
```

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    [JsonPropertyName("nrs")]
+    public string Name { get; set; }
+
+    [JsonPropertyName("ap1")]
+    public string FamilyName1 { get; set; }
```
abdallahbeshi, line 38: Surname?

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    [JsonPropertyName("ap2")]
+    public string FamiliyName2 { get; set; }
```
abdallahbeshi, line 41: SecondSurname?

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
```diff
@@ -0,0 +1,42 @@
...
+    public string FiscalYear { get; set; }
+}
+
+public class Inte
+{
+    [JsonPropertyName("nif")]
+    public string TaxPayerIdentificationNumber { get; set; }
+
+    [JsonPropertyName("nrs")]
+    public string Name { get; set; }
```
abdallahbeshi, line 35: if i remember correctly this is either the first name or the company name, pls double check
and if yes, i'd go with something like FirstNameOrCompanyName

src/Core/Mews.Fiscalizations.Core.Tests/Compression/GzipCompressionTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    private const string Message = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+
+    [Test]
+    public void GzipCompress_Decompress_Returns_Original_String()
+    {
+        Assert.DoesNotThrow(() =>
+        {
+            var compressedBytes = Message.Compress(encoding: System.Text.Encoding.ASCII);
+
+            var decompressedString = compressedBytes.Decompress(encoding: System.Text.Encoding.ASCII);
```
abdallahbeshi, line 17: import the using `using System.Text;`

abdallahbeshi (changes requested):

src/Basque/Mews.Fiscalizations.Basque/Dto/Bizkaia/Header/BizkaiaHeaderData.cs
inakiesteve: I think this is the FiscalYear the invoice belongs to so it might not be wise to hardcode it. While in most cases it should match the current year, there could be scenarios like in the first days of January where we could receive invoices from the previous year.

abdallahbeshi: Makes sense 👍

abdallahbeshi (approved):

PR #230, opened 2023-10-30
Upgrade the projects to use .NET 8 and C# 12

abdallahbeshi:
## Description

Upgrade the projects to use .NET 8 and C# 12

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [x] Is breaking change.
- [ ] Documentation updated.


inakiesteve (dismissed):

src/Spain/Mews.Fiscalizations.Spain/Converters/ModelToDtoConverter.cs
```diff
@@ -96,7 +96,7 @@ private SujetaType Convert(TaxSummary summary)
     {
         return new SujetaType
         {
-            Exenta = summary.TaxExempt.Map(items => items.Select(i => Convert(i)).ToArray()).GetOrNull(),
+            Exenta = summary.TaxExempt.Match(items => items.Select(i => Convert(i)).ToArray(), _ => null),
```
inakiesteve, line 99: Do we really need this? NoExenta just below uses Map so I would rather have the same behavior in both properties unless there is something else that I am missing? 😅

abdallahbeshi: Its because of the return type in the map/match, anyway i improved it further and got rid of the map and match.

inakiesteve (approved):

PR #263, opened 2025-02-19
Improved xml deserialization error handling in Basque fiscalization

inakiesteve:
## Description

Improved xml deserialization error handling in the Basque fiscalization when we receive badly formatted Xml responses from the Basque api.

Fixes #issue

## Type of change
- [x ] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Basque/Mews.Fiscalizations.Basque/DtoToModelConverter.cs
+++ b/src/Basque/Mews.Fiscalizations.Basque/DtoToModelConverter.cs
@@ -14,23 +14,6 @@ public static class DtoToModelConverter
         string tbaiIdentifier,
         String1To100 signatureValue)
     {
-        if (response is null)
-        {
-            var errorDescription = "Server error. Please try again.";
-            return new SendInvoiceResponse(
-                xmlRequestContent: xmlRequestContent,
-                xmlResponseContent: xmlResponseContent,
-                qrCodeUri: qrCodeUri,
-                tbaiIdentifier: tbaiIdentifier,
-                received: DateTime.UtcNow,
-                state: InvoiceState.Refused,
-                description: errorDescription,
-                signatureValue: signatureValue,
-                validationResults: [
-                    new SendInvoiceValidationResult(ErrorCode.ServerErrorTryAgain, errorDescription)
-                ]
-            );
-        }
         var result = response.Salida;
         return new SendInvoiceResponse(
             xmlRequestContent: xmlRequestContent,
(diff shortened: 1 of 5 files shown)


src/Basque/Mews.Fiscalizations.Basque/DtoToModelConverter.cs
```diff
@@ -14,6 +14,23 @@ internal static SendInvoiceResponse Convert(
         string tbaiIdentifier,
         String1To100 signatureValue)
     {
+        if (response is null)
```
abdallahbeshi, line 17: I would prefer to have this somewhere else and not in this method, or at least make the response nullable in the parameters.

src/Basque/Mews.Fiscalizations.Basque/DtoToModelConverter.cs
```diff
@@ -14,6 +14,23 @@ internal static SendInvoiceResponse Convert(
...
+        {
+            var errorDescription = "Server error. Please try again.";
+            return new SendInvoiceResponse(
+                xmlRequestContent: xmlRequestContent,
+                xmlResponseContent: xmlResponseContent,
+                qrCodeUri: qrCodeUri,
+                tbaiIdentifier: tbaiIdentifier,
+                received: DateTime.UtcNow,
+                state: InvoiceState.Refused,
```
abdallahbeshi, line 26: I don't think its correct for us to decide whether the invoice is refused or not, this should be only taken from the response.

also if the invoice is refused, i don't think we should be returning the qr code

src/Basque/Mews.Fiscalizations.Basque/TicketBaiClient.cs
```diff
@@ -123,15 +123,30 @@ private async Task<SendInvoiceResponse> SendTicketBaiInvoiceAsync(TicketBaiInvoi
...
-            qrCodeUri: invoiceData.QrCodeUri,
-            xmlRequestContent: signedRequest.OuterXml,
-            xmlResponseContent: responseContent,
-            tbaiIdentifier: invoiceData.TbaiIdentifier,
-            signatureValue: invoiceData.TrimmedSignature
-        );
+
+        try
+        {
+            var ticketBaiResponse = XmlSerializer.Deserialize<TicketBaiResponse>(responseContent);
```
abdallahbeshi, line 129: Did we check the technical documentation to see if they return a specific response format in case of a server error?

I think its best if we can do something like this but of course that depends if they return some specific format in case of a server error:
if (success) -> deserialize the response to TicketBaiResponse
if (error) -> deserialize the error response
if something is unhandled -> return some custom error response.

abdallahbeshi (changes requested):

abdallahbeshi (approved):

PR #264, opened 2025-02-21
Improve fiskaly error handling

inakiesteve:
## Description

Please include a summary of the change and which issue is fixed.

Fixes #issue

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [ ] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

### DIFF
--- a/src/Germany/Mews.Fiscalizations.Germany/V2/Model/Result/ErrorResult.cs
+++ b/src/Germany/Mews.Fiscalizations.Germany/V2/Model/Result/ErrorResult.cs
@@ -37,6 +37,11 @@ public sealed class ErrorResult
 
     private static FiskalyError MapError(Dto.FiskalyErrorResponse error)
     {
+        if (error is null)
+        {
+            return FiskalyError.InvalidResponse;
+        }
+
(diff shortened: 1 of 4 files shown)


src/Germany/Mews.Fiscalizations.Germany/V2/Model/Result/ErrorResult.cs
```diff
@@ -37,6 +37,11 @@ internal static ErrorResult MapException(JsonReaderException exception, string r
 
     private static FiskalyError MapError(Dto.FiskalyErrorResponse error)
     {
+        if (error is null)
+        {
+            return FiskalyError.InvalidResponse;
+        }
```
abdallahbeshi, line 43: This cannot be ever null based on its current usage, no? i believe it would crash on the deserialization and since there is a try/catch, it would use the MapException method and not this one

inakiesteve: Hey, I believe in this case the Json deserialization will just convert to a null object and not throw any exception so we could actually receive a null value

abdallahbeshi (approved):

PR #266, opened 2025-03-27
Switching from Srv4Pos to Infrasec

inakiesteve:
## Description

Introduced support to make requests for Swedish integration via infrasec (previously we were using Srv4Pos apis). Included some tests to create a ccu and check the status of it and also sending different types of transactions. 

Fixes #issue

## Type of change
- [ ] Bug fix.
- [ x] Feature.
- [ ] Consolidation.

## Checklist
- [ x] Is breaking change.
- [ ] Documentation updated.


abdallahbeshi (approved):
There is space for multiple improvements but for now we will publish the library and revisit this in the next few days.

PR #321, opened 2026-06-12
EXP-4396: support DSFinV-K cash point closing without transactions

andresgrobastrigo:
## What

Adds support for submitting a DSFinV-K `cashPointClosing` for a day **without transactions**.

Per the [fiskaly DSFinV-K API](https://developer.fiskaly.com/api/dsfinvk/v1#operation/insertCashPointClosing), on a day without transactions the `transactions` and `cash_statement` objects are no longer required and **must not be used**, and the head's `first/last_transaction_export_id` are `null`:

```json
{
  "head": {
    "first_transaction_export_id": null,
    "last_transaction_export_id": null,
    "export_creation_date": 1700000000
  },
  "client_id": "…",
  "cash_point_closing_export_id": 123
}
```

## Changes

- **`CashPointClosingMapper`** — when the closing has no transactions, emit `null` head ids and drop `transactions` + `cash_statement` (null is omitted by the serializer's `WhenWritingNull` default). Also tolerates a null `Transactions` collection.
- **`InsertCashPointClosingRequest`** — `first/last_transaction_export_id` now carry `[JsonIgnore(Condition = Never)]` so they serialize **explicitly as `null`** instead of being dropped by the global `WhenWritingNull` default.
- **Test** — new integration test `InsertCashPointClosingWithoutTransactionsSucceeds`.
- **Versions** — Fiskaly `2.4.0 → 2.5.0`, All `38.0.0 → 38.1.0`.

## Notes

The branching is done in the mapper, so the correct wire format is produced regardless of what a caller passes for the empty-day case (no consumer change required to get the fix).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Fiskaly/Mews.Fiscalizations.Fiskaly/DTOs/DSFinVK/CashPointClosings/InsertCashPointClosingRequest.cs
+++ b/src/Fiskaly/Mews.Fiscalizations.Fiskaly/DTOs/DSFinVK/CashPointClosings/InsertCashPointClosingRequest.cs
@@ -29,9 +29,6 @@ internal sealed class CashPointClosingHead
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string BusinessDate { get; init; }
 
-    // DSFinV-K expects these keys present even on a zero-transaction day, where both are null.
-    // JsonIgnoreCondition.Never overrides the serializer's global WhenWritingNull default so the
-    // null values are written explicitly rather than dropped.
     [JsonPropertyName("first_transaction_export_id")]
     [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
     public string FirstTransactionExportId { get; init; }
--- a/src/Fiskaly/Mews.Fiscalizations.Fiskaly/DTOs/DSFinVK/CashPointClosings/InsertCashPointClosingRequest.cs
+++ b/src/Fiskaly/Mews.Fiscalizations.Fiskaly/DTOs/DSFinVK/CashPointClosings/InsertCashPointClosingRequest.cs
@@ -29,11 +29,15 @@ internal sealed class CashPointClosingHead
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string BusinessDate { get; init; }
 
-    // Required by the spec (the caller sends "0" on zero-transaction days).
+    // DSFinV-K expects these keys present even on a zero-transaction day, where both are null.
+    // JsonIgnoreCondition.Never overrides the serializer's global WhenWritingNull default so the
+    // null values are written explicitly rather than dropped.
     [JsonPropertyName("first_transaction_export_id")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
     public string FirstTransactionExportId { get; init; }
 
     [JsonPropertyName("last_transaction_export_id")]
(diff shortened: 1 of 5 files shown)


src/Fiskaly/Mews.Fiscalizations.Fiskaly/Mappers/DSFinVK/CashPointClosings/CashPointClosingMapper.cs
```diff
@@ -43,11 +44,11 @@ public static InsertCashPointClosingRequestDto MapInsertRequest(CashPointClosing
...
                     CultureInfo.InvariantCulture
                 ),
-                FirstTransactionExportId = closing.FirstTransactionExportId,
-                LastTransactionExportId = closing.LastTransactionExportId,
+                FirstTransactionExportId = hasTransactions ? closing.FirstTransactionExportId : null,
+                LastTransactionExportId = hasTransactions ? closing.LastTransactionExportId : null,
             },
-            Transactions = transactions.Select(MapTransaction).ToList(),
-            CashStatement = MapCashStatement(closing.CashStatement),
+            Transactions = hasTransactions ? transactions.Select(MapTransaction).ToList() : null,
```
jairo-blanco-mews, line 50: By default we should return empty collections instead of null. Unless this is needed by business restrictions or something

src/Fiskaly/Mews.Fiscalizations.Fiskaly/DTOs/DSFinVK/CashPointClosings/InsertCashPointClosingRequest.cs
```diff
@@ -29,11 +29,12 @@ internal sealed class CashPointClosingHead
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string BusinessDate { get; init; }
 
-    // Required by the spec (the caller sends "0" on zero-transaction days).
     [JsonPropertyName("first_transaction_export_id")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
```
jairo-blanco-mews, line 33: Just curious, what is this meant for?

andresgrobastrigo: In this case we do need the param to be included in the Json request (per Fiskaly validations) but with explicit "null" value when it does not have it. It is the way to ensure json serializer includes it always. Reference: https://support.fiskaly.com/hc/en-001/articles/4961016737938-DSFinV-K-Is-it-possible-to-insert-a-cash-point-closing-without-any-transaction

src/Fiskaly/Mews.Fiscalizations.Fiskaly/Mappers/DSFinVK/CashPointClosings/CashPointClosingMapper.cs
andresgrobastrigo: Fiskaly documentation specifically reflects that transactions object is not expected when there are no actual transactions. In this way the validations on their side are not run. Reference: https://support.fiskaly.com/hc/en-001/articles/4961016737938-DSFinV-K-Is-it-possible-to-insert-a-cash-point-closing-without-any-transaction

jairo-blanco-mews (approved):

PR #325, opened 2026-09-16
EXP-4971: assert the cash register read carries back TSS and software version

andresgrobastrigo:
## Description

`CashRegisterTests.GetCashRegisterSucceeds` only asserted `IsSuccess` and the client id, so it would have passed even if the response dropped everything else. It now also asserts `TssId` and `SoftwareVersion`.

Those two fields are what a caller needs in order to tell an up to date cash register from a stale one, so this test is the only place where that read is exercised against Fiskaly for real. It backs [mews#88803](https://github.com/MewsSystems/mews/pull/88803), which reads the cash register before upserting it and compares exactly those two fields to decide whether the write is still needed. The upsert in the fixture's setup sends both, so a GET that does not echo them back would now fail here instead of silently breaking the comparison in the monolith.

Tests only. No production code, no library release, no package bump.

## Type of change
- [ ] Bug fix.
- [ ] Feature.
- [x] Consolidation.

## Checklist
- [ ] Is breaking change.
- [ ] Documentation updated.

`GetCashRegisterSucceeds`: the read returns the TSS id and software version that the setup's upsert sent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


moxly (approved):
## Review Summary

**Verdict: APPROVE**
The added assertions correctly verify the values supplied by the fixture setup.

---

[Open this review run in Navigator](https://navigator.mews.com/runs/code-review/f7de2015-d704-484c-8e83-c1f08a33a521)

